PR #19, opened 2015-05-08
fix marker title/subtitle

bsudekum:


RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -12,7 +12,7 @@ @implementation RCTMapboxGL
 
 @end
 
-@interface MGLAnnotation ()
+@interface addGLAnnotation ()
```
1ec5, line 15: This should be called something else, like `RCTMGLAnntation`. Classes in frameworks/components should always start with an uppercase prefix, either the project initials or your personal initials.

bsudekum, line 15: good to know, will change

PR #39, opened 2015-05-22
Remove reference to MGLMetricsLocationManager.h

1ec5:
MGLMetricsLocationManager.h was removed upstream in mapbox/mapbox-gl-native#1491.


bsudekum:
:+1:

PR #42, opened 2015-05-24
Add geocoding API

bsudekum:
Adds a geocoding function:

``` js
this.refs.mapRef.geocode('Chicago', true, (err, res) => {
  if(err) console.log(err);
  console.log(res)
})
```

Args:
- query, string
- proximity, bool
- callback

I'm not sold the this. My biggest complaint is that it differs from how the other methods work, IE it is not a mixin. Ideally methods and this function are implemented in the same manner. `geocode` does not need a `ref` because it is not actually manipulating the map. Perhaps in the future it would be a good idea to tack on a function which centers the map at the results from `geocode`. In this case, it will need a `ref` and called in the same way as mixin.

``` jsx
<Text style={styles.text} onPress={() => this.setDirectionAnimated(mapRef, 0)}>
  Set direction to 0
</Text>
<Text style={styles.text} onPress={() => this.refs.mapRef.geocode('Chicago', true, (err, res) => { if(err) console.log(err); console.log(res) })}>
geocode 'Chicago'
</Text>
```

### DIFF
--- a/RCTMapboxGL.ios.js
+++ b/RCTMapboxGL.ios.js
@@ -23,6 +23,11 @@ var MapView = React.createClass({
   statics: {
     Mixin: MapMixins
   },
+  getInitialState() {
+    return {
+      userLocation: null
+    }
+  },
   _onChange(event: Event) {
     if (!this.props.onRegionChange) {
       return;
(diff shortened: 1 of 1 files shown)


1ec5:
It does feel kind of odd that geocoding is associated with the map at all, since you might want to use it without even showing a map.

bsudekum:
@1ec5 true. The only reason I associated with a map is so I can grab the accessToken without making the user pass it in as an argument.

tmcw:
Same, this should be a module or another export rather than part of a component.

1ec5:
> The only reason I associated with a map is so I can grab the accessToken without making the user pass it in as an argument. 

Developers using Mapbox GL directly are now advised to set the access token globally in the Info.plist anyways, rather than through `+[MGLAccountManager setAccessToken:]`. Any React Native component can look at this Info.plist entry using `-[NSBundle objectForInfoDictionaryKey:]`.

tmcw:
Also this code should use the fetch api instead of polyfilled XmlHTTPRequest

bsudekum:
Rolling into a module.

PR #44, opened 2015-05-30
fix indentation, add annotation function

bsudekum:
ref: https://github.com/bsudekum/react-native-mapbox-gl/issues/43

Adds the ability to: 

``` jsx
<Text onPress={() => this.addAnnotation(mapRef, [{
  latitude: 0,
  longitude:  0,
  title: 'This is marker 1',
},{
  latitude: 20,
  longitude:  10,
  subtitle: 'foo'
}])}> // Adds new markers to the map without redrawing
```

My only issue here is that this blows away the previous markers on the map. I'm not certain if this a good thing or a bad thing. Perhaps marker management should be left up to the user. 

This also fixes indentation.


bsudekum:
Note, the api has changed from `addAnnotation` to `addAnnotations`.

PR #46, opened 2015-06-03
add select annotation method

bsudekum:
This adds `selectAnnotationAnimated(mapref, annotationPlaceInArray)` function which opens an annotation callout.

This changes how annotations are stored internally. I changed `annotations` from a dictionary to an array because of the way `selectAnnotationAnimated` works. The second argument is the annotations in the annotation array, IE pass in `0` and the first annotation's callout will open. This would not be possible with a dictionary as dictionaries are not ordered. 

@1ec5 a review here would be great.

### DIFF
--- a/RCTMapboxGL/RCTMapboxGL.m
+++ b/RCTMapboxGL/RCTMapboxGL.m
@@ -103,11 +103,8 @@ RCT_EXPORT_MODULE();
     if (_newAnnotations) {
         // Take into account any already placed pins
         if (_annotations.count) {
-            NSMutableArray *removeableKeys = [NSMutableArray array];
-            for (RCTMGLAnnotation *oldKey in _annotations){
-                [removeableKeys addObject:oldKey];
-                [_map removeAnnotation: oldKey];
-            }
+            [_map removeAnnotations: _annotations];
+            _annotations = nil;
         }
 
         _annotations = _newAnnotations;
@@ -192,7 +189,7 @@ RCT_EXPORT_MODULE();
                                                  @"magneticHeading": @(userLocation.heading.magneticHeading),
                                                  @"trueHeading": @(userLocation.heading.trueHeading),
                                                  @"isUpdating": [NSNumber numberWithBool:userLocation.isUpdating]} };
-  
+    
     [_eventDispatcher sendInputEventWithName:@"topLoadingFinish" body:event];
 }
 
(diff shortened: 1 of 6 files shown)


RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -102,36 +101,18 @@ - (void)setAnnotations:(NSMutableDictionary *)annotations
...
-                if (![newKeys containsObject:oldKey]){
-                    [removeableKeys addObject:oldKey];
-                }
-            }
-
-            // Remove each of the "removableKeys" from the map
-            if (removeableKeys.count){
-                NSArray *removed = [_annotations objectsForKeys:removeableKeys notFoundMarker:[NSNull null]];
-                [_map removeAnnotations: removed];
+            for (RCTMGLAnnotation *oldKey in _annotations){
```
1ec5: Instead of this `for` loop, call `[_map removeAnnotations:_annotations]`, then call `[_annotations removeAllObjects]`. (Or you can set `_annotations` to `nil` instead of emptying it – your call.)

RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -247,6 +228,17 @@ - (BOOL)mapView:(RCTMapboxGL *)mapView annotationCanShowCallout:(id <MGLAnnotati
     return YES;
 }
 
+- (void)selectAnnotationAnimated:(NSUInteger)annotationInArray
+{
+    if (_annotations) {
```
1ec5: If `-[RCTMapboxGL setAnnotations:]` is never called, `-selectAnnotationAnimated:` does nothing. Should it assert?

RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -247,6 +228,17 @@ - (BOOL)mapView:(RCTMapboxGL *)mapView annotationCanShowCallout:(id <MGLAnnotati
...
+        if ([_annotations count] <= annotationInArray) return NSLog(@"%s", "Annotation not found in annotations array");
```
1ec5: If the caller passes in a bogus ID, that’s programmer error, so assert (with `NSAssert`) instead of logging.

RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -94,7 +105,7 @@ - (dispatch_queue_t)methodQueue
         RCTMapboxGL *mapView = viewRegistry[reactTag];
         if([mapView isKindOfClass:[RCTMapboxGL class]]) {
             if ([annotations isKindOfClass:[NSArray class]]) {
-                NSMutableDictionary *pins = [NSMutableDictionary dictionary];
+                NSMutableArray* pins = [[NSMutableArray alloc] init];
```
1ec5: It’s easier to just call `[NSMutableArray array]`. (Also below.)

1ec5:
This will work for now, but as soon as you start to allow the developer to remove individual annotations, the idea of an “annotation index” will no longer work. You’ll eventually have to give each annotation a unique ID (it could still be an `NSUInteger`, but you can’t treat it as an array index anymore).

PR #47, opened 2015-06-04
add removeAnnotation method

bsudekum:
Adds `removeAnnotation` method.


1ec5:
Annotation indices are kind of pointless now because after you delete an annotation, the indices of all the annotations after it immediately change. So the client code has to maintain its own parallel array of annotations just to keep track of them.

PR #50, opened 2015-06-17
Add rightCalloutAccessory

bsudekum:
This updates the annotation prop to look like:

``` json
{
  "latitude": "required",
  "longitude":  "required",
  "title": "optional string",
  "subtitle": "optional string",
  "id": "optional string, unique identifier.",
  "rightCalloutAccessory": {
    "width": "required if url specified",
  }
}
```

Results in:

An image can be added via http or `image!yourImage.png` which follows react-natives image [api in place currently](https://facebook.github.io/react-native/docs/image.html#adding-static-resources-to-your-app-using-images-xcassets)

This also adds the event `onRightAnnotationTapped` which is fired when the rightcalloutaccessory is tapped. It returns the annotation object. 

Finally, I added the key `id` to the annotation object. This is a string and decided upon by the dev. I'm thinking it will come in handy when the user taps the accessory.

/cc @1ec5


brentvatne:
Nice one :smile:

PR #53, opened 2015-06-30
Update to work with react-native@0.7.0

bsudekum:
ref: https://github.com/bsudekum/react-native-mapbox-gl/issues/51

The final piece of the puzzle here is regarding `MGLMapboxMetricsEnabledSettingShownInApp`. With `mapbox-gl-native@0.4.0`, `MGLMapboxMetricsEnabledSettingShownInApp` is required if you do not provide analytics configuration. For now, I think it's okay to roll with this off and later down the line, expose this api.

@1ec5 I'm having a hard time getting this project to edit the plist.info of it's parent project. Is it possible for a child xcode project to edit the plist of the parent project? So far I've had no luck getting the correct path to the plist. Ideally, this would be added to this PR and would not require the dev to set this during the install process. 

Other changes in this pr:
- updates mapbox-gl-native to 0.4.0
  - you should see improved label rendering, [ref](https://www.mapbox.com/blog/mapbox-mobile-polylines/)
- changes how events are triggered and handled internally.
- Fixes an issue where `onRegionChange` and `onRegionWillChange` threw an error if one was not present
- Moves the example.md to to example.jsx to made it easier to read


1ec5:
> Is it possible for a child xcode project to edit the plist of the parent project?

No, because libRCTMapboxGL.a is prebuilt and is a dependency of the app target. Mapbox GL comes with two primary installation workflows: [manual installation inside Xcode](https://www.mapbox.com/mapbox-gl-ios/#binary) and [automatic installation via CocoaPods](https://www.mapbox.com/mapbox-gl-ios/#cocoapods). The [setup steps for this React project](https://github.com/bsudekum/react-native-mapbox-gl/blob/master/install.md) are based on the manual installation process, which is primarily intended for developers of existing projects who are comfortable mucking around with Xcode settings and don’t need the handholding that CocoaPods provides.

Our intent is that setting `MGLMapboxMetricsEnabledSettingShownInApp` be a deliberate action on the part of the developer that comes after implementing custom UI for toggling Mapbox Metrics. For most developers, [adding a Settings bundle](https://www.mapbox.com/guides/first-steps-gl-ios/#settings) that contains the Mapbox Metrics switch would be less work. Requiring them to set this flag manually in the project editor shouldn’t be a special burden, either, because they’ll have to set their access token and [enable Location Services](https://www.mapbox.com/guides/first-steps-gl-ios/#enabling-location-services) right there anyways.

1ec5:
Since this React project expects `npm` to be used as the dependency manager, could you customize `npm install` to add the Settings bundle to any project already in the target directory?

PR #54, opened 2015-07-01
Access requireNativeComponent through public module interface

brentvatne:
As per https://github.com/facebook/react-native/issues/1821 - with 0.7.0+, requiring of any React Native internal modules will need to go through the public interface :smile: :+1:

### DIFF
--- a/RCTMapboxGL.ios.js
+++ b/RCTMapboxGL.ios.js
@@ -1,8 +1,7 @@
 'use strict';
 
 var React = require('react-native');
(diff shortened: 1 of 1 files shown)


bsudekum:
@brentvatne great, are there any changes you see that need to happen for 0.7.0+? Working on a release for 0.6.0 now but could potential make a release for both.

brentvatne:
Nothing else stands out to me! This fix will work on both 0.6 and 0.7 :smile:

PR #65, opened 2015-07-04
add emptMemoryCache method

bsudekum:
Native documentation found here: https://www.mapbox.com/mapbox-gl-ios/api/Classes/MGLMapView.html#//api/name/emptyMemoryCache

/cc @1ec5


1ec5:
When would you expect React Native developers to need to call this method? I filed mapbox/mapbox-gl-native#1833 because I’m not sure it should even be public in the first place.

1ec5:
Sounds like this method isn’t meant to be public. mapbox/mapbox-gl-native#1833 tracks making it visible only to iosapp.

Issue #27, opened 2015-05-10
Add all params that are in mapbox gl


bsudekum:
They should have identical APIs. ref: https://www.mapbox.com/mapbox-gl-ios/api/Classes/MGLMapView.html#properties

manglav:
can you also add this property as well? That will help me get started. I can submit a PR too, I just don't know objective C.
visibleCoordinateBounds

PR #67, opened 2015-07-09
Adding scrollEnabled and zoomEnabled props

VonD:
A small step towards #27


bsudekum:
Awesome @VonD. My only feedback is that if `scrollEnabled={true}` and/or `zoomEnabled={true}` is not present in your render function, it defaults to `false`. Can you set both of these to true in the [`getDefaultProps`](https://github.com/mapbox/react-native-mapbox-gl/blob/master/RCTMapboxGL.ios.js#L84-L88) function?

``` js
getDefaultProps() {
    return {
      styleURL: 'asset://styles/mapbox-streets-v7.json',
      scrollEnabled: true,
      zoomEnabled: true
    };
  }
```

VonD:
@bsudekum done. Sorry, I should have thought about it.

bsudekum:
:+1:

PR #86, opened 2015-09-06
Delay adding annotations to the map view

bsudekum:
I am seeing more timing issues when adding a custom image to an annotation or rightCalloutAccessory when running on device.

The error is very nondescript, but giving the annotations more time before adding to the map view solved the issue:

I think some of this might be attributed to https://github.com/mapbox/mapbox-gl-native/issues/1675

/cc @1ec5

### DIFF
--- a/RCTMapboxGL/RCTMapboxGL.h
+++ b/RCTMapboxGL/RCTMapboxGL.h
@@ -17,7 +17,7 @@
 - (instancetype)initWithEventDispatcher:(RCTEventDispatcher *)eventDispatcher;
 
 - (void)setAccessToken:(NSString *)accessToken;
-- (void)setAnnotations:(NSArray *)annotations;
+- (void)setAnnotations:(NSMutableArray *)annotations;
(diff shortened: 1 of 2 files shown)


RCTMapboxGL/RCTMapboxGL.h
```diff
@@ -17,7 +17,7 @@
 - (instancetype)initWithEventDispatcher:(RCTEventDispatcher *)eventDispatcher;
 
 - (void)setAccessToken:(NSString *)accessToken;
-- (void)setAnnotations:(NSArray *)annotations;
+- (void)setAnnotations:(NSMutableArray *)annotations;
```
1ec5, line 20: This change is unnecessary, and you didn’t pair it with a change in the .m anyways.

1ec5:
Can you build with the symbols version of the Mapbox GL library? The screenshot above shows an unsymbolicated stack, which is useless in tracking down the problem.

bsudekum:
closing in favor of https://github.com/mapbox/react-native-mapbox-gl/pull/89

PR #95, opened 2015-10-07
Add polyline and polygon API

bsudekum:
Took another stab at https://github.com/mapbox/react-native-mapbox-gl/pull/66, this time I was way more successful! Some notes:
- This will be a `1.0.0` release due to the breaking changes around annotations
- `type` is now required on each object in the annotations array. Enum values `point`, `polyline` and `polygon` are valid
- The annotations object will now look like:

``` json
[{
  "coordinates": "required. For type polyline and polygon must be an array of arrays. For type point, single array",
  "type": "required: point, polyline or polygon",
  "title": "optional string. Only used for type=point",
  "subtitle": "optional string. Only used for type=point",
  "fillAlpha": "optional, only used for type=polygon. Controls the opacity of polygon",
  "fillColor": "optional string hex color including #, only used for type=polygon",
  "strokeAlpha": "optional number from 0-1. Only used for type=poyline. Controls opacity of line",
  "strokeColor": "optional string hex color including #, used for type=polygon and type=polyline",
  "strokeWidth": "optional number. Only used for type=poyline. Controls line width",
  "id": "optional string, unique identifier",
  "rightCalloutAccessory": {
    "width": "required if url specified"
  },
  "annotationImage": {
    "width": "required if url specified"
  }
}]
```

@1ec5 a review would be :100:

### DIFF
--- a/RCTMapboxGL/RCTMapboxGL.m
+++ b/RCTMapboxGL/RCTMapboxGL.m
@@ -136,7 +136,13 @@ RCT_EXPORT_MODULE();
 
 - (UIColor *)mapView:(MGLMapView *)mapView strokeColorForShapeAnnotation:(RCTMGLAnnotationPolyline *)shape
 {
-    return [self getUIColorObjectFromHexString:shape.strokeColor alpha:1];
+    if ([shape isKindOfClass:[RCTMGLAnnotationPolyline class]]) {
+        return [self getUIColorObjectFromHexString:shape.strokeColor alpha:1];
+    } else if ([shape isKindOfClass:[RCTMGLAnnotationPolygon class]]) {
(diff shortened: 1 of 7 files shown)


RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -105,29 +104,64 @@ - (void)layoutSubviews
 - (void)setAnnotations:(NSMutableArray *)annotations
 {
     _newAnnotations = annotations;
-    [self performSelector:@selector(updateAnnotations) withObject:nil afterDelay:0.1];
+    [self performSelector:@selector(updateAnnotations) withObject:nil afterDelay:0.5];
```
1ec5, line 107: Why the increased delay?

PR #106, opened 2015-10-16
Updates steps explaining header_search_paths

dsibiski:
The React Native module is actually 1 more level up in the folder structure now since the `RCTMapboxGL.xcodeproj` is in the nested `ios` folder.


bsudekum:
Thanks!

PR #123, opened 2015-10-31
Add CocoaPods spec file to simplify project setup.

christoph-jerolimov:
For example this fix the setup requirement that the Xcode header search path must be changed anymore / and anytime the node_modules folder was resetted.

See changed ios/installation for more details.

### DIFF
--- a/ios/install.md
+++ b/ios/install.md
@@ -1,7 +1,20 @@
 # Installation Process
 
 Notes:
+
 * `react-native ^v0.4.3` is required
+
+## Use CocoaPods
+
+If you already use [CocoaPods](https://cocoapods.org/) in your react-native project you add the native
+part of Mapbox GL with just one line of code:
+
+1. `npm install react-native-mapbox-gl --save`
+1. Add `pod 'RCTMapboxGL', :path => 'node_modules/react-native-mapbox-gl/ios'` to your `Podspec` file and re-run `pod install`.
+
+See also the react-native [0.13 release notes](https://github.com/facebook/react-native/releases/tag/v0.13.0)
+
+## Manually
 
 1. `npm install react-native-mapbox-gl --save`
 1. In the XCode's `Project navigator`, right click on project's name ➜ `Add Files to <...>` ![](https://cldup.com/k0oJwOUKPN.png)
--- a/ios/RCTMapboxGL.podspec
+++ b/ios/RCTMapboxGL.podspec
@@ -1,0 +1,25 @@
+Pod::Spec.new do |s|
+  s.name                = "RCTMapboxGL"
+  s.version             = "1.0.0"
+  s.summary             = "A Mapbox GL react native module for creating custom maps."
+  s.homepage            = "https://github.com/mapbox/react-native-mapbox-gl#readme"
+  s.license             = "BSD"
+  s.author              = "Bobby Sudekum"
+  s.source              = { :git => "https://github.com/mapbox/react-native-mapbox-gl.git", :tag => "#{s.version}" }
+  s.platform            = :ios, "7.0"
+
+  s.source_files        = "RCTMapboxGL/RCTMapboxGL.{h,m}", "RCTMapboxGL/RCTMapboxGLManager.{h,m}"
+  s.dependency            "Mapbox-iOS-SDK", "2.1.2"
+
+  # These resources, library, frameworks and libraries spec is already part of
+  # the Mapbox-iOS-SDK dependency above. The npm `preinstall` hock download
+  # them automatically. Instead of the dependency we could also use the
+  # following attributes to use the local files...
+  # I keep this it it makes development easier of fix somes "troubles" later.
+
+#  s.resources           = "RCTMapboxGL/Mapbox.bundle", "RCTMapboxGL/Settings.bundle"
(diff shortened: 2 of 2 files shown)


bsudekum:
This looks good. Going to test this weekend.

christoph-jerolimov:
FYI, In my project I added the Podfile within the `ios` folder with a **clean** Xcode project and CocoaPods.

My `AppDelegate`:

``` objc
#import <React/RCTRootView.h>

...
    NSURL *mainBundle = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];

    // Just override the bundle URL, use debug env later,
    mainBundle = [NSURL URLWithString:@"http://localhost:8081/index.ios.bundle?platform=ios&dev=true"];

    RCTRootView *rootView = [[RCTRootView alloc] initWithBundleURL:mainBundle
                                                        moduleName:@"AppName"
                                                 initialProperties:nil
                                                     launchOptions:launchOptions];

    self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
    UIViewController *rootViewController = [[UIViewController alloc] init];
    rootViewController.view = rootView;
    self.window.rootViewController = rootViewController;
    [self.window makeKeyAndVisible];
```

My `Podfile`:

``` ruby
source 'https://github.com/CocoaPods/Specs.git'

pod 'React', :path => '../node_modules/react-native'
pod 'React/RCTGeolocation', :path => '../node_modules/react-native'
pod 'React/RCTImage', :path => '../node_modules/react-native'
pod 'React/RCTNetwork', :path => '../node_modules/react-native'
pod 'React/RCTText', :path => '../node_modules/react-native'
pod 'React/RCTWebSocket', :path => '../node_modules/react-native'

#pod 'RCTMapboxGL', :path => '../node_modules/react-native-mapbox-gl/ios' # npm version
pod 'RCTMapboxGL', :path => '../../react-native-mapbox-gl/ios' # local clone
```

Maybe I will create a blog posts about this later.

bsudekum:
I just released 1.2.0. I'm going to save this for the next release to better triage bug reports.

nevir:
Awesome, thanks for the cocoapods setup, it'll make it much easier to consume this library!

chris-erickson:
That setup phase was rough, can't wait for this! :chocolate_bar:

edvinasbartkus:
:+1: awesome!

chris-erickson:
@bsudekum When can we expect the next release?

bsudekum:
@jerolimov how will this effect someone who has an existing project and they'll adding this module? Will they have to use the `project.xcworkspace` file now? I'm not well versed in cocoapod workflows. 

/cc @incanus

christoph-jerolimov:
No, this effect the users only if they explicit import the spec file in their Podfile. So there is no risk to merge it. If you want use the "react-native / old way" and modify the project yourself it will work.

With the spec file you have the option to use Cocoapods. CocoaPods will create a new Xcode project called Pods which is generated and just import this Pods project into your Project. This makes updates really easy.

Both works parallel without trouble and any user can choice what they use in their project.

(I'm only unsure if we should move and rename the file into the top folder, `react-native-mapbox-gl/react-native-mapbox-gl.podspec` instead of `react-native-mapbox-gl/ios/RCTMapboxGL.podspec`.)

Should I create a demo project?

jerolimov:
@bsudekum Maybe you can talk to Justin R. Miller and Minh Nguyễn (didn't know their username here) which maintain the `Mapbox-iOS-SDK` project on cocoapods: https://cocoapods.org/pods/Mapbox-iOS-SDK

christoph-jerolimov:
@bsudekum Maybe you can talk to @incanus and @1ec5 which maintain already the `Mapbox-iOS-SDK` project on cocoapods: https://cocoapods.org/pods/Mapbox-iOS-SDK

ios/install.md
```diff
@@ -1,8 +1,21 @@
...
+
 * `react-native ^v0.4.3` is required
 
+## Use CocoaPods
+
+If you already use [CocoaPods](https://cocoapods.org/) in your react-native project you add the native
+part of Mapbox GL with just one line of code:
+
+1. `npm install react-native-mapbox-gl --save`
+1. Add `pod 'RCTMapboxGL', :path => 'node_modules/react-native-mapbox-gl/ios'` to your `Podspec` file and re-run `pod install`.
```
1ec5, line 13: By the way, this points to a local, npm-based installation of RCTMapboxGL. I guess it isn’t strictly necessary to install the pod from cocoapods.org.

1ec5:
@bsudekum, as @jerolimov says, there’s no impact for someone who wants to follow the existing installation workflow. This would be more convenient for someone who’s already using CocoaPods for other modules. I haven’t tried the pod locally, but the Podfile looks good to me.

ios/install.md
christoph-jerolimov, line 13: Your right, publishing it on CocoaPods is not required and maybe redundant.

The release notes of react-native 0.13 contains some infos how facebook will handle this. And my own tries shows the same problems.

https://github.com/facebook/react-native/releases/tag/v0.13.0

christoph-jerolimov:
Thanks @bsudekum :+1:

Issue #118, opened 2015-10-26
NPM failed install, example does not work in react-native v0.13.0


syarul:
On Windows,
`message: `MapView` has no propType for native prop `RCTMapbox.scaleX` of native type `number``

bsudekum:
I'm going to need a little more information than what was provided to know what the real issue is. Is there an error in the xcode console? Did you follow ever install step? 

> NPM failed install

What failed?

> `RCTMapbox.scaleX`

Are you adding a prop that is not supposed to be there? `scaleX` is found no where in this project.

syarul:
All exact step install for android, nothing being added it's just pure copy of example.js. The issue is the error is ios specifix, but I'm running android on windows, probably breaking changes in v0.13

kanerogers:
I'm having the same issue too. :frowning:

bsudekum:
Confirmed, this is an error for rn 0.13.0+

kanerogers:
FWIW, I was able to get things working by adding the following `PropType`s to `index.android.js`:

``` javascript
{
    scaleY: React.PropTypes.number,
    scaleX: React.PropTypes.number,
    translateY: React.PropTypes.number,
    translateX: React.PropTypes.number,
    rotation: React.PropTypes.number
}
```

Happy days! :cat:

bsudekum:
Very strange... these props have nothing to do with this module...

kanerogers:
Yeah. I'm thinking that as of 0.13 RN is now passing these attributes down
to every native component.. or something.

bsudekum:
@brentvatne have you seen this issue? As a quick fix, I'm going to add the props above to fix this issue.

syarul:
+1 @kanerogers

syarul:
Working in v0.14.1, though to get mapbox installed on windows need to pass `--ignore-scripts` to disable ios script that would make npm 3 installed failed on android. I'll submit PR for this

`npm install react-native-mapbox-gl --save --ignore-scripts`

If your screen turn out black or white give a minutes or 2 to see the maps loaded afterwards.

fixed issue #107

PR #129, opened 2015-11-11
add small fix to readme for npm 3 failed installation on windows

syarul:
Fixed Issue #118


bsudekum:
True, the post script is not needed for android. Thanks.

PR #132, opened 2015-11-11
Set deployment target to 7.0

bsudekum:
I don't think we'll be supporting 7.x here, too much overhead.


christoph-jerolimov:
Supporting 7.x would be great and I expect no troubles because react-native as well as Mapbox-iOS-SDK supports 7.x, too. :smiley:

bsudekum:
Per conversation with @1ec5, 7.0 should be fine.

christoph-jerolimov:
Nice :+1:

1ec5:
To clarify, supporting older versions of iOS can result in some overhead, especially whenever the system UI changes dramatically or the Location Services permissions model changes. But that’s something for mapbox-gl-native to worry about, and we aren’t at the point where we’d need to drop iOS 7.0 support from the Objective-C code. Going forward, react-native-mapbox-gl should just track whatever mapbox-gl-native uses as a minimum deployment target.

Issue #120, opened 2015-10-27
no map after install


dannytce:
I've tried the example and got this: http://prntscr.com/8vx1yo

There also Warning in console: Warning: Failed propType: typeChecker is not a function Check the render method of `Example`.

All the features work just fine. But i can't see streets, etc.

bsudekum:
Interesting, if you change the style to `light-v8` or `dark-v8`, does the map show up?

dannytce:
Yes, that did the trick. Also `satellite-v8` is working fine. But there is the same problem with `emerald-v8`.

syarul:
what's wrong with streets-v8 most of the time it doesn't seem to work?

onchainguy-btc:
Same issue for me :-(

bsudekum:
This is an upstream issue and will be solved when [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native) 3.0.0 is released.

bsudekum:
Check out v2.0.0, this issue should be fixed in it.

priithaamer:
I'm still getting this warning with Mapbox 2.2.0 and react-native 0.16, but only when setting annotations on `<Mapbox annotations={[...]} />` component. When annotations is empty array or entirely missing, the warning is gone.

Now since that yellow warning box was introduced in react-native 0.16 this warning has become very annyoing :) Fortunately, it can be turned off with `console.disableYellowBox = true;` though.

PR #140, opened 2015-11-17
[iOS] Updates mapbox-gl-native to v3.0.0pre7

bsudekum:
This:
- `ENABLE_BITCODE = YES;`
- Fixes marker race condition
- Fixes an issue where maps would not be shown randomly https://github.com/mapbox/react-native-mapbox-gl/issues/120

/cc @1ec5, not sure if this is the most elegant way to handle the styleurl change


bsudekum:
On second thought, I might just retire `asset://` and require `mapbox://` on all style urls.

bsudekum:
@1ec5 are there real advantages to `[MGLStyle lightStyleURL];`?

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -72,7 +72,21 @@ - (void)updateMap
         _map.scrollEnabled = _scrollEnabled;
         _map.zoomEnabled = _zoomEnabled;
         _map.showsUserLocation = _showsUserLocation;
-        _map.styleURL = _styleURL;
+        if ([_styleURL  isEqual: @"lightStyleURL"]) {
```
1ec5: Instead of accepting magic strings here, there should be a separate JavaScript object that vends string constants that can be set as the styleURL.

1ec5: If you do that, styleURL can remain an NSURL property, which it should be.

1ec5:
Do not retire mapbox: URLs: that's how developers set custom styles. Let the iOS SDK decide what is and isn't a valid style URL.

The advantage of lightStyleURL et al. is that developers won't have to keep changing the URL string as we release new versions of these styles in the style API. Besides, it's always good programming practice to use constants (or factory methods in this case) instead of hard-coded literals.

MGLStyle is the beginning of a full-fledged runtime style API that will be implemented in coming releases, so it should be treated as an object in its own right.

ios/RCTMapboxGL/RCTMapboxGL.m
1ec5: If the separate factory object idea doesn't fit too well in this project's architecture, there could be a standalone function that maps default style name strings like `"emerald"` to the corresponding mapbox: URLs. The developer would call that function and pass the return value into the `styleURL` property.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -43,6 +43,20 @@ - (NSArray *)customDirectEventTypes
...
+- (NSDictionary *)constantsToExport
+{
+    return @{
+             @"styles": @{
+                 @"light": [[MGLStyle lightStyleURL] absoluteString],
+                 @"dark": [[MGLStyle darkStyleURL] absoluteString],
+                 @"streets": [[MGLStyle streetsStyleURL] absoluteString],
+                 @"emerald": [[MGLStyle emeraldStyleURL] absoluteString],
+                 @"satellite": [[MGLStyle satelliteStyleURL] absoluteString],
+                 @"satelliteHybrid": [[MGLStyle hybridStyleURL] absoluteString],
```
1ec5: We were planning to call this style just Hybrid in iOS SDK documentation, since that’s what Google and Apple call it.

1ec5:
This is looking mighty good.

Issue #142, opened 2015-11-18
Set the center to user location


gundam-wing:
How can I set the center of the map to the coordinates of the user's location?

priithaamer:
Currently, i was only able to do it in a slightly hacky way but it kind of works with something like this:

``` javascript
constructor() {
  super();
  this.state = {centered: false};
}

handleUpdateUserLocation(center) {
  if (this.state.centered === false) {
    this.setCenterCoordinateAnimated('map', center.latitude, center.longitude);
    this.setState({centered: true});
    // You may also want to turn off user location tracking here
  }
}

render() {
  return (<MapBox
    ref="map"
    onUpdateUserLocation={this.handleUpdateUserLocation.bind(this)}
  />);
}
```

friedbunny:
It looks like we're lacking access to the `userTrackingMode` setting on iOS, which would allow the map to automatically recenter on user location updates — see #145.

gundam-wing:
@priithaamer Thanks

bsudekum:
You can now use `userTrackingMode={this.userTrackingMode.follow}` in v2.0.0 and it will center on the user.

doriansgithub:
i can't get it to work, even with the follow tracking mode.  I can see the my location beacon, but it will not center and/or zoom on it's own. 

- (void)zoomToUserLocationMapBox{
    NSLog(@"zoomToUsersLocationMapBox");
    
    _mapBoxView.userTrackingMode = MGLUserTrackingModeFollowWithHeading;
    _mapBoxView.showsUserLocation = YES;
   _mapBoxView.showsHeading = YES;
    
    CLLocationCoordinate2D centerCoordinate =
    CLLocationCoordinate2DMake(38.894368, -77.036487);
    
    [_mapBoxView setCenterCoordinate:centerCoordinate
                           zoomLevel:15
                            animated:YES];
}

I also don't understand why the sample provides these random coordinates, instead of the user's location code.

doriansgithub:
It now works when I do a search off a code that has nothing to do with Mapbox.  There must be something linked between Mapbox and MapKit that's triggering it.  I'll find it.

PR #149, opened 2015-11-21
Add tracking modes, update to pre10

### LINKED ISSUE #145: iOS: support `userTrackingMode`
The iOS SDK offers the `mapView.userTrackingMode` property to allow the map view to automatically center on and follow the user's coordinates. It looks like Android has support here via [`UserLocationTrackingMode`](https://github.com/mapbox/react-native-mapbox-gl/blob/3a7109b6ed0c3982dfc3b2cec100fa16f77e185e/index.android.js#L29), but the iOS implementation is missing.

[Here are the possible values](https://github.com/mapbox/mapbox-gl-native/blob/ddc4a8193e775d9cd9280ad7d9bb6975b4a323ee/include/mbgl/ios/MGLTypes.h#L17-L27) for `MGLMapView.userTrackingMode`:

``` objc
/** The mode used to track the user location on the map. */
typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
    /** The map does not follow the user location. */
    MGLUserTrackingModeNone              = 0,
    /** The map follows the user location. */
    MGLUserTrackingModeFollow,
    /** The map follows the user location and rotates when the heading changes. */
    MGLUserTrackingModeFollowWithHeading,
    /** The map follows the user location and rotates when the course changes. */
    MGLUserTrackingModeFollowWithCourse,
};
```

/cc @bsudekum

bsudekum:
Adds user tracking modes, closes https://github.com/mapbox/react-native-mapbox-gl/issues/145 and https://github.com/mapbox/react-native-mapbox-gl/issues/142

@1ec5 @incanus I'm seeing an issue with `MGLUserTrackingModeFollowWithCourse` and `MGLUserTrackingModeFollowWithHeading`, namely that they don't work. `follow` seems to work fine but, even when I hard code `_map.userTrackingMode` to with course or heading, the map does not pan to the users location nor follow it with heading. Have you seen this before?


1ec5:
Are you testing in the Simulator? The Simulator doesn’t support heading simulation.

bsudekum:
Ah yep

bsudekum:
@1ec5 hmm on device, same issue, heading and course do not work... even when hard coded.

incanus:
Another common thing that can turn this off is any sort of programmatic or gestural movement of the map. The best way to figure out how it’s happening is to set a breakpoint in the setter for the mode and then see what triggers it to turn off, or if it’s truly just not being set in the first place for some reason.

bsudekum:
Although not ideal, I added a timeout to setting the tracking mode. This seems to do the trick for now, I'm going to revisit this later.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -46,16 +46,24 @@ - (NSArray *)customDirectEventTypes
...
+             @"mapStyles": @{
                  @"light": [[MGLStyle lightStyleURL] absoluteString],
                  @"dark": [[MGLStyle darkStyleURL] absoluteString],
                  @"streets": [[MGLStyle streetsStyleURL] absoluteString],
                  @"emerald": [[MGLStyle emeraldStyleURL] absoluteString],
                  @"satellite": [[MGLStyle satelliteStyleURL] absoluteString],
                  @"hybrid": [[MGLStyle hybridStyleURL] absoluteString],
+            },
+             @"userTrackingMode": @{
+                     @"none": [NSNumber numberWithUnsignedInt:MGLUserTrackingModeNone],
```
1ec5, line 58: These can be written as simply `@(MGLUserTrackingModeNone)`.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -217,6 +219,16 @@ - (void)setStyleURL:(NSURL *)styleURL
     [self updateMap];
 }
 
+- (void)setUserTrackingMode:(int)userTrackingMode
+{
+    if (userTrackingMode > 3 || userTrackingMode < 0) {
+        _userTrackingMode = 0;
+    } else {
+        _userTrackingMode = userTrackingMode;
```
1ec5, line 227: Should issue an error and return in this case.

1ec5:
I’m guessing this is the reason heading or course tracking doesn’t activate when you attempt to enable them synchronously: as soon as you try to switch to heading or course tracking, [this line](https://github.com/mapbox/mapbox-gl-native/blob/b3dbdb389ff712ace00857e3e72894f0bb3afe3a/platform/ios/MGLMapView.mm#L2660) resets the mode to no tracking if the map view has yet to obtain a valid user location. Effectively, this means that you need to have entered follow mode at least once before entering heading or course tracking mode.

iosapp [shows the user dot _after_ setting the tracking mode](https://github.com/mapbox/mapbox-gl-native/blob/8a0bce780847546c161670420e4d0f1e3a6cbdca/ios/app/MBXViewController.mm#L115). By a fit of Objective-C luck, the map view’s `userLocation` is therefore `nil`, so `userLocation.coordinate` is (0, 0), which is a valid coordinate.

We ported this check from the legacy iOS SDK, where it was introduced in mapbox/mapbox-ios-sdk-legacy@f781914dcc2727e9ab974f2f257c4ee02d38b86d. @incanus, is this behavior still relevant?

PR #171, opened 2015-12-05
Add onTap and onLongPress

### LINKED ISSUE #108: Capture touch on map to add marker at said coordinates
Is there a way to catch a user tap or even dragging the marker to certain coordinates?

If the user taps it would create the annotation there?

bsudekum:
Adds an `onTap` and `onLongPress` (1 second) event listener.

closes: https://github.com/mapbox/react-native-mapbox-gl/issues/108

/cc @1ec5 for :eyes:

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -90,6 +90,19 @@ RCT_EXPORT_MODULE();
     _map = [[MGLMapView alloc] initWithFrame:self.bounds];
     _map.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
     _map.delegate = self;
+    
+    UITapGestureRecognizer *doubleTap = [[UITapGestureRecognizer alloc] initWithTarget:self action:nil];
+    doubleTap.numberOfTapsRequired = 2;
+    [_map addGestureRecognizer:doubleTap];
+    
+    UITapGestureRecognizer *singleTap = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleSingleTap:)];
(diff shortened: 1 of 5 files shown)


ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -404,6 +422,36 @@ - (MGLAnnotationImage *)mapView:(MGLMapView *)mapView imageForAnnotation:(id<MGL
...
 }
 
+- (void)handleSingleTap:(UITapGestureRecognizer *)tap
+{
+    CLLocationCoordinate2D location = [_map convertPoint:[tap locationInView:_map] toCoordinateFromView:_map];
+    
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"latitude": @(location.latitude),
+                                     @"longitude": @(location.longitude),
```
1ec5, line 428: Maybe it would be worth sending along the screen coordinates (x, y) too, in case the developer wants to position some UI element over the map at the tap point. Or you could break out methods like `-convertPoint:toCoordinateFromView:` into separate methods that bridge directly between JS and MGLMapView. That could let developers do more with this functionality besides just handling taps.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -218,6 +231,11 @@ - (void)setStyleURL:(NSURL *)styleURL
     _styleURL = styleURL;
     [self updateMap];
 }
+- (void)setAttributionButtonVisibility:(BOOL)isVisible
```
1ec5: Are you planning to expose this property to JS?

bsudekum: whoops that snuck in. This will be in another pr

bsudekum:
:+1: added screen coordinates of the tap events.

PR #172, opened 2015-12-05
Remove ontap

bsudekum:
The onTap function seems to be clobbering all tap functions on the screen. This means you can no longer tap annotations to trigger the info window opening. @1ec5 I'm inclined to remove the api all together, is there any way around this?

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -54,9 +54,6 @@ var MapView = React.createClass({
   _onUpdateUserLocation(event: Event) {
     if (this.props.onUpdateUserLocation) this.props.onUpdateUserLocation(event.nativeEvent.src);
   },
-  _onTap(event: Event) {
-    if (this.props.onTap) this.props.onTap(event.nativeEvent.src);
-  },
   _onLongPress(event: Event) {
     if (this.props.onLongPress) this.props.onLongPress(event.nativeEvent.src);
   },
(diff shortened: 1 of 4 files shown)


1ec5:
You might try setting the gesture recognizer’s `cancelsTouchesInView` to `NO` and see if that works. Otherwise, yeah, I would implement the older `-touchesBegan:withEvent:` method, which lets you work with raw touches rather than complete gestures.

If you do get it working, note that your events will duplicate the ones MGLMapView responds to. There’s no good way to coordinate the gesture recognizers you provide with the ones internal to MGLMapView. That’s why iosapp only adds a long press gesture recognizer that doesn’t conflict. So I think it’d be fine to eliminate this gesture recognizer because its behavior would be somewhat counterintuitive.

bsudekum:
Could not get it to work, going to merge this and revisit the problem later.

xxsnakerxx:
@bsudekum While I'm have no time to create pull requests (because I don't know how to create them properly) you can add features which you need from my fork, e.g. onTap https://github.com/xxsnakerxx/react-native-mapbox-gl/blob/master/ios/RCTMapboxGL/RCTMapboxGLManager.m#L54-L73 https://github.com/xxsnakerxx/react-native-mapbox-gl/blob/master/ios/RCTMapboxGL/RCTMapboxGLManager.m#L826-L859

Issue #175, opened 2015-12-06
Android demo doesn't work


Otouto:
Trying to run [default demo](https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/example.js) and get an error:

![image](https://cloud.githubusercontent.com/assets/1306915/11615471/e4349564-9c69-11e5-9ca8-111caca00e57.png)

Otouto:
#177

bsudekum:
This has been fixed and released

PR #176, opened 2015-12-06
adjust RCTMapbox initialization

Otouto:
#175 fix android demo


bsudekum:
Gosh, hate to see this list pile up. @Otouto do you know if there is an associated ticket in rn core?

Otouto:
@bsudekum unfortunately no. Found out which `propTypes` should be added through experimental path.

PR #178, opened 2015-12-07
Add attribution, logo and compass toggle props

bsudekum:
Allows users to toggle on and off the attribution button, logo and compass.

/cc @1ec5

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -96,6 +96,9 @@ var MapView = React.createClass({
         url: React.PropTypes.string
       })
     })),
+    attributionButtonIsHidden: React.PropTypes.bool,
+    logoIsHidden: React.PropTypes.bool,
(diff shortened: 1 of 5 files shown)


index.ios.js
```diff
@@ -115,7 +118,10 @@ var MapView = React.createClass({
       showsUserLocation: false,
       styleUrl: this.Mixin.mapStyles.streets,
       zoomEnabled: true,
-      zoomLevel: 0
+      zoomLevel: 0,
+      attributionButtonIsHidden: false,
+      logoIsHidden: false,
+      compassIsHidden: false
```
1ec5, line 124: It’d be less confusing to spin these properties in a positive light, e.g. `showsLogo`. (See `showsUserLocation` above.)

PR #181, opened 2015-12-08
Add more event listeners

bsudekum:
- `onFinishLoadingMap`
- `onStartLoadingMap`
- `onLocateUserFailed`

/cc @1ec5

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -469,6 +469,39 @@ RCT_EXPORT_MODULE();
     }
 }
 
+- (void)mapViewDidFinishLoadingMap:(MGLMapView *)mapView
+{
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"loaded": @YES
+                                     }
+                             };
+    
+    [_eventDispatcher sendInputEventWithName:@"onFinishLoadingMap" body:event];
+}
+- (void)mapViewWillStartLoadingMap:(MGLMapView *)mapView
+{
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"loading": @YES
+                                     }
+                             };
+    
+    [_eventDispatcher sendInputEventWithName:@"onStartLoadingMap" body:event];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFailToLocateUserWithError:(NSError *)error
+{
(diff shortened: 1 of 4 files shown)


ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -469,6 +469,39 @@ - (void)handleLongPress:(UITapGestureRecognizer *)sender
     }
 }
 
+- (void)mapViewDidFinishLoadingMap:(MGLMapView *)mapView
+{
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"loaded": @YES
```
1ec5: Is there any point to this `loaded` property if it’s always `YES`?

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -469,6 +469,39 @@ - (void)handleLongPress:(UITapGestureRecognizer *)sender
...
+                             };
+    
+    [_eventDispatcher sendInputEventWithName:@"onStartLoadingMap" body:event];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFailToLocateUserWithError:(NSError *)error
+{
+    NSDictionary *event = @{ @"target": mapView.reactTag,
+                             @"src": @{
+                                     @"code": @(error.code),
```
1ec5: `code` is one of the enum values [listed here](https://developer.apple.com/library/mac/documentation/CoreLocation/Reference/CoreLocationConstantsRef/#//apple_ref/c/tdef/CLError). Are developers expected to compare the code to (say) `2` to discover a network issue, or does React Native automatically expose the `CLError` values to the JS side as well?

ios/RCTMapboxGL/RCTMapboxGL.m
bsudekum: Yeah, these should be removed.

ios/RCTMapboxGL/RCTMapboxGL.m
bsudekum: True, do you think `message` is enough?

1ec5: `description` is intended for debugging purposes; you should never compare it to a string to identify a particular condition. The wording could also potentially change from release to release. An NSError object should be identified by its `code`, but that means you may have to export the CLError values as constants, just as you did for style URLs. Also, you should also send along the error’s `localizedDescription`: this is the potentially localized error message that’s suitable for presenting to the user.

PR #185, opened 2015-12-12
[Android] [Not Ready] Modify map state after initialization

bsudekum:
This begins the work that is necessary for updating the map state after initialization. This work will mainly live in https://github.com/mapbox/react-native-mapbox-gl/commit/d7172580282d1ff52b000b82b76f6862ab6cbd46. Still not sure if this is the correct pattern to follow, but it seems to be working well so far.

/cc @tobrun @bleege


android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
```diff
@@ -45,9 +58,9 @@ public String getName() {
 
     @Override
     public MapView createViewInstance(ThemedReactContext context) {
-        MapView mv = new MapView(context, "pk.foo");
-        mv.onCreate(null);
-        return mv;
+        mapView = new MapView(context, "pk.foo");
+        mapView.onCreate(null);
```
bleege, line 76: @bsudekum Looks like you may want to set a style here via `mapView.setStyle(Style.MAPBOX_STREETS)` for example.

bleege:
One other thing to note is that it'll need to call all the lifecycle methods as documented on the [Examples page](https://www.mapbox.com/android-sdk/examples/).  Perhaps this is already done elsewhere though as my current knowledge of ReactNative is limited.

christopherdro:
I get a black map when using prop `styleUrl={'asset://styles/streets-v8.json'}`. Removing it defaults to street view.

bsudekum:
@bleege added constants with https://github.com/mapbox/react-native-mapbox-gl/commit/bb129c396ad3ac8e4913c96ed156898cd548eb6f#diff-511703cd599f72abb07f0ddde993736fR54.

I'm also trying to add custom icons, but for some reason the first `annotationImage` is styling all annotations on the map. Anything here https://github.com/mapbox/react-native-mapbox-gl/pull/185/files#diff-16f337bda10b484bae5e45d7c47f1fa2R130 look funny to you?

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
```diff
@@ -59,6 +88,22 @@ public void setAccessToken(MapView view, @Nullable String value) {
...
+        HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+        connection.connect();
+        InputStream input = connection.getInputStream();
+
+        x = BitmapFactory.decodeStream(input);
+        return new BitmapDrawable(x);
+    }
+
     @ReactProp(name = PROP_ANNOTATIONS)
     public void setAnnotations(MapView view, @Nullable ReadableArray value) {
```
christopherdro: We should remove all annotations when this prop is called. If not, anytime the state of annotations changes, the previous state markers stay visible.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
```diff
@@ -59,6 +88,22 @@ public void setAccessToken(MapView view, @Nullable String value) {
...
+
+        HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+        connection.connect();
+        InputStream input = connection.getInputStream();
+
+        x = BitmapFactory.decodeStream(input);
+        return new BitmapDrawable(x);
+    }
+
     @ReactProp(name = PROP_ANNOTATIONS)
```
christopherdro, line 107: We should remove all annotations when this prop is called. If not, anytime the state of annotations changes, the previous state markers stay visible.

christopherdro:
@bsudekum annotationImage isn't working for me at all. Defaults to standard map marker.

bsudekum:
@christopherdro this is a work in progress, see https://github.com/mapbox/react-native-mapbox-gl/pull/185#issuecomment-165563405

Right now it will only work if annotationImage is present on the first annotation in the array. Not quite sure why.

christopherdro:
@bsudekum For sure, I was just trying to reference that comment. We are thinking about using mapbox for our project so I will be doing my best to contribute.

tobrun:
@bsudekum @christopherdro 
The annotation issue is something tracked here https://github.com/mapbox/mapbox-gl-native/issues/3317. I'm going to take a look at that, note that the fix will introduce a breaking api change.

bsudekum:
@tobrun [this](https://github.com/mapbox/mapbox-gl-native/issues/3317#issuecomment-165826112) did the trick! @christopherdro want to give this a pull and check it out?

christopherdro:
@bsudekum Nice! Working for me.

bsudekum:
I'm keen to merge this soon as it adds some good additions. Still missing pieces that can be added down the line without breaking changes:
- resizing annotation images
- clicking popups

christopherdro:
Any thoughts on the comment above regarding clearing markers on state changes?

This morning I came across an issue with Android where map keeps snapping back to its original position every time time I try to scroll.

I tested it on the following devices.

Galaxy x5 (5.0) - Works
Nexus 4 (5.1.1) - Broken
Nexus 5x (6) - Broken
Moto X Pure (5.1.1) Broken

Will have a chance a bit later to dig in more.

bsudekum:
@christopherdro can you double check `userTrackingMode` is `none`? Which commenting are referring to?

christopherdro:
Hmm, good call! Looks like the syntax changed a bit.

Whats strange though is that I was testing the same code side by side.

Heres the comment from earlier. https://github.com/mapbox/react-native-mapbox-gl/pull/185#discussion_r47980222

bsudekum:
Yeah it looks like: https://github.com/mapbox/react-native-mapbox-gl/pull/185/files#diff-90cc9c6e35fdb83c539a6c366af8f28bR106

`userTrackingMode={this.userTrackingMode.none}`

christopherdro:
@bsudekum btw, you ever on discord? I tried pinging you the other day.

PR #187, opened 2015-12-13
implemented pass map bounds to onRegionChange and onFinishLoadingMap …

22rostislav:
Implemented pass map bounds to onRegionChange and onFinishLoadingMap events.

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -325,11 +325,13 @@ RCT_EXPORT_MODULE();
 {
 
     CLLocationCoordinate2D region = _map.centerCoordinate;
+    NSDictionary *bounds = [self getBounds];
 
     NSDictionary *event = @{ @"target": self.reactTag,
                              @"src": @{ @"latitude": @(region.latitude),
                                         @"longitude": @(region.longitude),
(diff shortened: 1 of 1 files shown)


bsudekum:
@22rostislav I think it'd make more sense to create a function to get the bounds rather than send it through on these two methods as they seem unrelated to me. Want to reroll this and create something like `getVisibleCoordinateBounds`?

brianjd:
Passing map bounds to onRegionChange makes sense to me.  I had been looking for this, thought I was missing something in documents.   Why doesn't it make sense to pass the bounds of the "new region" to onRegionChange?

xxsnakerxx:
+1 `getVisibleCoordinateBounds` make sense as additional API because it will be async. It will make code more complex if you want get current center and bounds when region was changed

brianjd:
It does makes sense to add getVisibleCoordinateBounds, as you may be interested in getting bounds outside of onRegionChange.   But at the same time if you are monitoring onRegionChange, you are probably interested in the region bounds, and considering center coords are being passed, why not pass the bounds as well?   Just my reasoning.

bsudekum:
I'd like to keep the `onRegionChange` function as light as possible as overloading this listener can be detrimental to panning performance. I think adding a method to get current visible bounds is the best path forward.

PR #198, opened 2015-12-19
Use ids for selecting annotations

bsudekum:
This changes:
- Annotations will not be wiped away when new annotations are added to map
- remove and selecting an annotation is done with an id vs a place in an array
- adds `removeAllAnnotations` function

/cc @christopherdro

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.h
+++ b/ios/RCTMapboxGL/RCTMapboxGL.h
@@ -34,7 +34,7 @@
 - (void)setCenterCoordinateAnimated:(CLLocationCoordinate2D)coordinates;
 - (void)setCenterCoordinateZoomLevelAnimated:(CLLocationCoordinate2D)coordinates zoomLevel:(double)zoomLevel;
 - (void)selectAnnotationAnimated:(NSString*)selectedId;
-- (void)removeAnnotation:(NSString*)selectedId;
+- (void)removeAnnotation:(NSString*)selectedIdentifier;
 - (void)removeAllAnnotations;
 - (void)setVisibleCoordinateBounds:(MGLCoordinateBounds)bounds edgePadding:(UIEdgeInsets)padding animated:(BOOL)animated;
 - (void)setAttributionButtonVisibility:(BOOL)isVisible;
--- a/ios/RCTMapboxGL/RCTMapboxGL.h
+++ b/ios/RCTMapboxGL/RCTMapboxGL.h
@@ -33,8 +33,9 @@
 - (void)setDirectionAnimated:(int)heading;
 - (void)setCenterCoordinateAnimated:(CLLocationCoordinate2D)coordinates;
 - (void)setCenterCoordinateZoomLevelAnimated:(CLLocationCoordinate2D)coordinates zoomLevel:(double)zoomLevel;
-- (void)selectAnnotationAnimated:(NSUInteger)annotationInArray;
-- (void)removeAnnotation:(NSUInteger)annotationInArray;
+- (void)selectAnnotationAnimated:(NSString*)selectedId;
+- (void)removeAnnotation:(NSString*)selectedId;
+- (void)removeAllAnnotations;
 - (void)setVisibleCoordinateBounds:(MGLCoordinateBounds)bounds edgePadding:(UIEdgeInsets)padding animated:(BOOL)animated;
 - (void)setAttributionButtonVisibility:(BOOL)isVisible;
 - (void)setLogoVisibility:(BOOL)isVisible;
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -22,7 +22,6 @@
     /* Map properties */
     NSString *_accessToken;
     NSMutableArray *_annotations;
-    NSMutableArray *_newAnnotations;
     CLLocationCoordinate2D _centerCoordinate;
     BOOL _clipsToBounds;
     BOOL _debugActive;
@@ -49,7 +48,7 @@ RCT_EXPORT_MODULE();
         _eventDispatcher = eventDispatcher;
         _clipsToBounds = YES;
         _finishedLoading = NO;
-        _newAnnotations = [NSMutableArray array];
+        _annotations = [NSMutableArray array];
     }
 
     return self;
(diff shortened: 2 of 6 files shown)


bsudekum:
@1ec5 how do these changes look to you?

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -117,23 +116,13 @@ - (void)layoutSubviews
...
-        }
-
-        _annotations = _newAnnotations;
-        for (int i = 0; i < [_newAnnotations count]; i++) {
-            [_map addAnnotation:_newAnnotations[i]];
-        }
+- (void)updateAnnotations:(NSMutableArray *) annotations {
+    for (int i = 0; i < [annotations count]; i++) {
+        [_annotations addObject:annotations[i]];
+        [_map addAnnotation:annotations[i]];
```
1ec5: It’s more efficient to call `-[MGLMapView addAnnotations:]`, passing in the entire array of annotations at once.

bsudekum: Nice, good call.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -117,23 +116,13 @@ - (void)layoutSubviews
 
 - (void)setAnnotations:(NSMutableArray *)annotations
 {
-    _newAnnotations = annotations;
-    [self performSelector:@selector(updateAnnotations) withObject:nil afterDelay:0.5];
+    [self performSelector:@selector(updateAnnotations:) withObject:annotations afterDelay:0.5];
```
1ec5, line 119: This method should blow away any existing annotations. Also, is the delay needed anymore? This is the only method that calls `-updatedAnnotations:`. Ideally that method should be inlined here.

bsudekum, line 119: Unfortunately the delay is still necessary. without it, no annotations.

1ec5:
Instead of having each annotation maintain its own ID, it would be far more efficient for RCTMapboxGL to keep an NSMutableDictionary that maps IDs to annotation objects. In fact, there’s no reason for RCTMapboxGL’s `_annotations` ivar to be an array: as long as you [add the annotations in the order they’re provided](https://github.com/mapbox/react-native-mapbox-gl/pull/198#discussion_r48097866), the order after that point no longer matters to RCTMapboxGL. So turn `_annotations` into an NSMutableDictionary and you can remove most of the for loops here.

ios/RCTMapboxGL/RCTMapboxGL.m
1ec5, line 119: That suggests that `-setAnnotations:` itself should perhaps be called on a delay. If you call `-setAnnotations:` long after setting up the view, a delay probably wouldn’t make sense – unless you’re also resetting the style on the same iteration of the run loop?

ios/RCTMapboxGL/RCTMapboxGL.m
1ec5: `-[MGLMapView setAnnotations:]` handles both deleting existing annotations and adding all the new ones.

ios/RCTMapboxGL/RCTMapboxGL.h
```diff
@@ -33,8 +33,9 @@
 - (void)setDirectionAnimated:(int)heading;
 - (void)setCenterCoordinateAnimated:(CLLocationCoordinate2D)coordinates;
 - (void)setCenterCoordinateZoomLevelAnimated:(CLLocationCoordinate2D)coordinates zoomLevel:(double)zoomLevel;
-- (void)selectAnnotationAnimated:(NSUInteger)annotationInArray;
-- (void)removeAnnotation:(NSUInteger)annotationInArray;
+- (void)selectAnnotationAnimated:(NSString*)selectedId;
```
1ec5, line 36: The prevailing style in Objective-C is to capitalize “ID”, or better yet, to spell it out as “identifier”. That’s because `id` is a keyword with special meaning in this language.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -358,27 +351,23 @@ - (BOOL)mapView:(RCTMapboxGL *)mapView annotationCanShowCallout:(id <MGLAnnotati
...
-    if ([_annotations count] != 0) {
-        [_map removeAnnotation:_annotations[annotationInArray]];
-        [_annotations removeObjectAtIndex:annotationInArray];
+    [_map removeAnnotation:[_annotations objectForKey:selectedIdentifier]];
+    [_annotations removeObjectForKey:[_annotations objectForKey:selectedIdentifier]];
+}
+
+- (void)removeAllAnnotations
+{
+    for(id key in _annotations) {
```
1ec5: You can replace this loop with a single call to `[_map removeAnnotations:_map.annotations]`.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -117,24 +116,19 @@ - (void)layoutSubviews
...
-        _annotations = _newAnnotations;
-        for (int i = 0; i < [_newAnnotations count]; i++) {
-            [_map addAnnotation:_newAnnotations[i]];
+- (void)updateAnnotations:(NSMutableArray *) annotations {
+    for (int i = 0; i < [annotations count]; i++) {
+        NSString *id = [annotations[i] id];
+        if ([id length] != 0) {
+            [_annotations setObject:[annotations objectAtIndex:i] forKey:id];
+        } else {
+            [_annotations setObject:[annotations objectAtIndex:i] forKey:[NSString stringWithFormat:@"id-%d", i]];
```
1ec5: Keys can be anything: you might as well make the key an NSNumber in this case, like this: `@(i)`. That way no one can accidentally collide with it by giving an annotation an ID of `id-0` or whatever.

Another option would be to make the key required (`nonnull`).

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -117,24 +116,19 @@ - (void)layoutSubviews
...
-                [_map removeAnnotations: _annotations];
-            }
-            _annotations = nil;
-        }
-
-        _annotations = _newAnnotations;
-        for (int i = 0; i < [_newAnnotations count]; i++) {
-            [_map addAnnotation:_newAnnotations[i]];
+- (void)updateAnnotations:(NSMutableArray *) annotations {
+    for (int i = 0; i < [annotations count]; i++) {
```
1ec5: It would potentially be a lot more efficient to write `for (RCTMGLAnnotation *annotation in annotations)` instead of repeated calls to `-objectAtIndex:`.

PR #211, opened 2015-12-26
not add existing on map annotation

22rostislav:
If annotation exists on map skip it


bsudekum:
@22rostislav can you tell me more about the issue this is fixing? Are you using the same id for multiple annotations?

christopherdro:
I think what we want here is to update a existing annotation any of its keys have been changed.

tmcw:
I don't think this patch should be accepted: updateAnnotations should do what it says on the tin and update all annotations, including ones that are currently on the map. Mutating annotations in-place goes against the React approach to one-way data.

PR #221, opened 2016-01-05
Support attribution and logo hide on Android

priithaamer:
Adds `attributionButtonIsHidden` and `logoIsHidden` property support for Android. They behave the same way as in iOS

Changes are reflected in API documentation.

Please let me know if it should be solved in some other way.


bsudekum:
This looks good. Can we change the rotation prop to be `rotationEnabled` across both?

priithaamer:
Ah, sorry, i only wanted to pull the first commit about attribution but github included my second commit as well. I will use branches next time.

However, i see that `rotateEnabled` is being used in Android and iOS SDK so i thought it would make sense to keep the same name in react-native library as well.

bsudekum:
@priithaamer can you rebase this off master? I went and fixed the rotation issue here https://github.com/mapbox/react-native-mapbox-gl/pull/230

bsudekum:
Added this in https://github.com/mapbox/react-native-mapbox-gl/commit/780fe619b3b3857d9186100f0584eb27f5f3dae7

PR #223, opened 2016-01-05
Make rotation prop consistent across platforms.

christopherdro:
This update changes the iOS prop to match Android's `rotationEnabled`.
Docs updated to reflect changes.


bsudekum:
I think we should keep as many breaking changes out of ios and default these to android. Let's go with `rotateEnabled` so we do not have to change https://github.com/mapbox/react-native-mapbox-gl/blob/master/ios/RCTMapboxGL/RCTMapboxGLManager.m#L77. Queuing up this change now.

PR #228, opened 2016-01-06
Add helper functions to convert annotations objects

christopherdro:
- Creates helper files converting annotations objects to MGLAnnotation.
- Add method for adding singe annotations. Exposed as a react method and used in loops with `addAnnotations` and the export view prop `annotations`.
- Add method for updating single a annotation.
- Fixes bug with removing annotations. [here](https://github.com/mapbox/react-native-mapbox-gl/pull/228/files#diff-28b9bb3037372337e6be55e0754f802dL363)


bsudekum:
This looks good. Will annotate code with a few changes.

index.ios.js
```diff
@@ -16,9 +16,15 @@ var MapMixins = {
   setCenterCoordinateZoomLevelAnimated(mapRef, latitude, longitude, zoomLevel) {
     NativeModules.MapboxGLManager.setCenterCoordinateZoomLevelAnimated(React.findNodeHandle(this.refs[mapRef]), latitude, longitude, zoomLevel);
   },
+  addAnnotation(mapRef, annotation) {
+    NativeModules.MapboxGLManager.addAnnotation(React.findNodeHandle(this.refs[mapRef]), annotation);
+  },
```
bsudekum: Let's remove this method. No need to have `addAnnotation` and `addAnnotations`

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -208,6 +244,17 @@ - (NSDictionary *)constantsToExport
     }];
 }
 
+RCT_EXPORT_METHOD(addAnnotation:(nonnull NSNumber *) reactTag
```
bsudekum: Remove, ref https://github.com/mapbox/react-native-mapbox-gl/pull/228/files#r49216047

bsudekum:
This PR should also update:
- example.js
- API.md - specifically explain that if the annotation id does not exist and `updateAnnotation` is called, it will still be added to the map. Also, explain that the annotation object passed in `updateAnnotation` will used the id inside of it update the annotation.

bsudekum:
:+1:

christopherdro:
:rocket:

PR #232, opened 2016-01-09
Respect height and width settings on annotationImage property

AidenMontgomery:
Currently images custom annotation images are not resized when being added to the map on Android.
This change uses the spriteFactory method that resizes the images before setting them on the annotation.


bsudekum:
Nice this looks good. Can you make the if statement check for both height and width?

AidenMontgomery:
I have added that now

bsudekum:
Looks good, thanks!

PR #235, opened 2016-01-11
Fixing set annotations to remove all prior annotations before adding …

mishkinf:
Fixing set annotations to remove all prior annotations before adding new annotations. According the the documentation 'Note, this will remove all previous annotations from the map' which was not the observed behavior

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -115,6 +115,7 @@ RCT_EXPORT_MODULE();
 
 - (void)setAnnotations:(NSMutableArray *)annotations
 {
+		[self removeAllAnnotations];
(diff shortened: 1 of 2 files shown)


ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -115,6 +115,7 @@ - (void)layoutSubviews
 
 - (void)setAnnotations:(NSMutableArray *)annotations
 {
+		[self removeAllAnnotations];
```
bsudekum: This has already been fixed in https://github.com/mapbox/react-native-mapbox-gl/pull/225

mishkinf:
Reverting the iOS portion of the fix

bsudekum:
:+1:

PR #259, opened 2016-01-28
Upgrade to 3.1.0

bsudekum:
This upgrades the Mapbox ios sdk up to v.3.1.0. v.3.1.0 moves to a dynamic library so now the install process is a bit different.

This also adds `contentInset` used for changing the center point of the mapview.

I'll also need to update the cocoa pods install before merging.

/cc @1ec5


ios/RCTMapboxGL/RCTMapboxGL.h
```diff
@@ -44,6 +43,7 @@
 - (double)zoomLevel;
 - (double)direction;
 - (CLLocationCoordinate2D)centerCoordinate;
+- (void)setContentInset:(UIEdgeInsets)inset;
```
1ec5: Any reason for not exposing a `-contentInset` getter? (You could declare both at one time with `@property UIEdgeInsets inset;`.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -81,6 +82,7 @@ - (void)updateMap
         [_map.attributionButton setHidden:_attributionButton];
         [_map.logoView setHidden:_logo];
         [_map.compassView setHidden:_compass];
+        _map.contentInset = _inset;
```
1ec5: I’d mildly recommend setting this before setting `userTrackingMode`, just to reduce the likelihood of edge cases, since both properties can cause the map to move.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -133,6 +133,16 @@ - (NSDictionary *)constantsToExport
     }
 }
 
+RCT_CUSTOM_VIEW_PROPERTY(contentInset, UIEdgeInsetsMake, RCTMapboxGL)
+{
+    int top = [json[0] intValue];
```
1ec5: Insets can be fractional, so you should take `doubleValue` if possible.

1ec5: If you declare these variables as `CGFloat`s and set them using `doubleValue` instead of `intValue`, the casts below go away.

ios/install.md
```diff
@@ -1,21 +1,21 @@
...
-8. Click on the `RCTMapboxGL` project. Under the `Build Settings` tab, search for `header_search_path`. Make sure `$(SRCROOT)/../../../React` and `$(SRCROOT)/../../react-native/React` are added and set to `recursive`. ![](https://cldup.com/81zUEHaKoX.png)
-9. You can now `require('react-native-mapbox-gl')` and build.
+_Make sure to follow these directions carefully._
+
+First,
+```bash
+npm install react-native-mapbox-gl --save
+```
+
+1. In the XCode's `Project navigator`, right click on project's name ➜ `Add Files to <...>`. Add `node_modules/react-native-mapbox-gl/ios/RCTMapboxGL.xcodeproj` ![](https://cldup.com/DTD2UZMYu5.png)
```
1ec5: Nit: It’s Xcode, not XCode.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -381,6 +383,12 @@ - (void)removeAnnotation:(NSString*)selectedIdentifier
     }
 }
 
+- (void) setContentInset:(UIEdgeInsets)inset
+{
+    _inset = inset;
```
1ec5: This class already has a `contentInset` property and thus a `_contentInset` ivar. You should update `_contentInset` and get rid of the `_inset` ivar (also above when you construct the map).

scripts/download-mapbox-gl-native-ios.sh
```diff
@@ -8,20 +8,9 @@ cd ios/
...
-    mkdir temp
-fi
-
-if [ -d ./RCTMapboxGL/Mapbox.bundle ]; then
-    rm -r RCTMapboxGL/Mapbox.bundle
-fi
-
-mkdir RCTMapboxGL/Mapbox.bundle
 
 curl -sS https://mapbox.s3.amazonaws.com/mapbox-gl-native/ios/builds/mapbox-ios-sdk-$VERSION.zip > temp.zip
```
1ec5: Since you’re only using the dynamic framework, make sure it’s installing the `-dynamic.zip` package so the download size is much smaller.

PR #265, opened 2016-02-01
Add v3.2.0

bsudekum:
/cc @1ec5


1ec5:
You might want to incorporate more of the [mapbox-gl-native changelog](https://github.com/mapbox/mapbox-gl-native/blob/master/CHANGELOG.md#ios-master). There have been changes that aren’t symbol changes but nonetheless impact the user experience, particularly with gesture recognition in user tracking mode.

I think this changelog should also point out that any existing installation of the component may need to be uninstalled first: the developer might not want all those extra framework linkages cluttering up the project.

Issue #266, opened 2016-02-02
iOS: Part of polyline is plotted on the map that shouldn't be


jabm111:
On iOS, if I plot a polyline between two points that are pretty close together, a whole other part of the polyline is plotted where it shouldn't be:

![screen shot 2016-02-02 at 4 32 38 pm](https://cloud.githubusercontent.com/assets/10445424/12765749/112c2c44-c9ce-11e5-8aa8-3394cc4153e4.png)

I works as expected on Android:

![screen shot 2016-02-02 at 4 32 55 pm](https://cloud.githubusercontent.com/assets/10445424/12765761/21732e68-c9ce-11e5-8c92-1d308cfced9e.png)

iOS and Android index files are identical: 

https://gist.github.com/jabm111/d54b01bdce8e4df40405

my package.json: 

https://gist.github.com/jabm111/05b25463f30c0e5f7997

(if I use RN 0.19.0 I get a black screen on Android like #264)

bsudekum:
Confirmed, coming from https://github.com/mapbox/react-native-mapbox-gl/blob/master/ios/RCTMapboxGL/RCTMapboxGLManager.m#L481

@1ec5 is there a way to rewrite this without `malloc`?

1ec5:
> is there a way to rewrite this without `malloc`?

No, there isn’t. However, you did find the bug: you must never reference `coord` after calling `free()` on it. Referencing a dangling pointer causes corruption issues like the one seen here, and it’s also a security hole. To fix this issue, you’ll need to store the return value of `-[RCTMGLAnnotationPolyline polylineAnnotation:strokeAlpha:strokeColor:strokeWidth:id:type:count:]` in a local variable, then call `free()`, then return that return value.

PR #268, opened 2016-02-08
Dont reference coord after freeing it

bsudekum:
Fixes: https://github.com/mapbox/react-native-mapbox-gl/issues/266

/cc @1ec5 for review


1ec5:
:+1:

PR #278, opened 2016-02-18
Replace all instances of styleUrl with styleURL. Fixes #273

### LINKED ISSUE #273: StyleUrl is inconsistent on Android and iOS (styleURL vs styleUrl)
So...
It appears that sometime recently iOS started using `styleURL` while Android and iOS I believe were previously both using `styleUrl`.
Not sure if Android is still using the latter version, but the current examples on the README demonstrate the inconsistency.

Don't think the inconsistency is intentional, but it led me to about an hour or so of testing.
I couldn't figure out why my map was a blank white screen in iOS.

Thanks, keep up the good work.

tmcw:
cc @bsudekum for the review


bsudekum:
:+1:. I think this spelling mistake warrants a dot release :confused:

Issue #279, opened 2016-02-18
Rotation for custom marker image


aksonov:
Any documentation for it? All I see is example polygon and custom image. Could i pass react native view or something like that?

tmcw:
You can add lines, polygons, and image-based markers to the map. Sorry, passing react native views is not supported.

aksonov:
@tmcw thanks, is it possible to rotate custom image? Would you accept PR to allow set angle for image?

tmcw:
@bsudekum rotation of point annotations would be possible if we exposed the underlying gl-native method, right?

aksonov:
Looks like I need to change `- (MGLAnnotationImage *)mapView:(MGLMapView *)mapView imageForAnnotation:(id<MGLAnnotation>)annotation`

`image = [UIImage imageNamed:localImagePath];`

rotate that image if not 'image!' prefix passed but something like 'rotatedImage!_degree_!

aksonov:
RCTMapBoxGL.m  file

aksonov:
I've added PR for this.

PR #280, opened 2016-02-18
support rotation of local annotation images

aksonov:
PR for #279, syntax of url should be "rotatedImage!_degree_!_image_name_"


bsudekum:
I think this change should be made upstream in mapbox-gl-native instead of hacking it in this module. ref https://github.com/mapbox/mapbox-gl-native/issues/2528

Issue #270, opened 2016-02-09
Could not build module 'Mapbox' (MGLAccountManager.h file not found)


munkius:
I followed the steps in the installation guide file, but I get the error the Mapbox framework could not be found. It get this after adding `libRCTMapboxGL.a` to the section `Link Binaries With Libraries`. Before doing this I can successfully build my project.

I'm working with react-native-mapbox-gl version 3.2.1.

![image](https://cloud.githubusercontent.com/assets/194593/12921947/d06eac36-cf4f-11e5-894c-edb959d82f2f.png)

JonnyBGod:
Same Error

bsudekum:
Could you verify:
1. `Mapbox.framework` is listed in `Embedded Binaries`
2. `Linked frameworks and Libraries` has `libRCTMapboxGL.a`

![](https://cldup.com/cOZRZt2jZG.png)

Uploading a screenshot just like this would help

ferndopolis:
Same error on 3.2.1. When I search for `MGLAccountManaer.h` in my project root

```
$ find . -type f -name "MGLAccountManager.h"                                                             
 ./node_modules/react-native-mapbox-gl/ios/Mapbox.framework/Mapbox.framework/Headers/MGLAccountManager.h
```

there is a Mapbox.framework nested inside Mapbox.framework.  Is that right?

munkius:
![image](https://cloud.githubusercontent.com/assets/194593/12984447/0e63a0b0-d0ee-11e5-9a1d-fed896b04c20.png)

bsudekum:
Hmmm that looks like. @ferndopolis something is up with your install process, when I run 

``` bash
find . -type f -name "MGLAccountManager.h"  
./node_modules/react-native-mapbox-gl/ios/Mapbox.framework/Headers/MGLAccountManager.h
```

Unsure how it would put the framework inside of the framework...

@munkius are you upgrading from a previous version?

munkius:
@bsudekum It's a clean install, what makes you think that?

bsudekum:
@munkius just wondering. An older version could have been messing with the build.

munkius:
Tried it a bunch of times with a fres workspace to be sure not to disturb you guys with a nonsense issue :) Hope this can work out, I'm really curious about getting Mapbox to work within react native.

bsudekum:
@munkius Here is a barebones setup: https://cloudup.com/c9J25rX9Jla

[...]

PR #283, opened 2016-02-18
Fix install process

bsudekum:
Fixes: https://github.com/mapbox/react-native-mapbox-gl/issues/270

Two issues were solved here:

``` bash
if [ -f ./Mapbox.framework ]; then
  echo "Old Mapbox.framework found. Removing it and installing a $VERSION"
  rm -rf ./Mapbox.framework
 fi
```

A `.framework` is a directory, not a file. This should have been `-d`. This resulted in nested `Mapbox.framework`.

The second issue this solves is around `MGLAccountManager.h`. `MGLAccountManager.h` was in the `.npmignore` file and somehow was being omitted from the `Mapbox Module`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -30,18 +30,6 @@ node_modules/
 npm-debug.log
 
 # project specific
-libMapboxGL.a
-MapboxGL.bundle/
-MapboxGL.h
-MGLAnnotation.h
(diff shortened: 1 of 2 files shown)


bsudekum:
Tested with a fresh install, works.

PR #300, opened 2016-03-07
Added method for getting current bounds [android]

maqen:
Using recommended approach until more native call is available

https://github.com/mapbox/mapbox-gl-native/issues/3863#issuecomment-181825298


bsudekum:
:+1: tested, works great. Thanks @maqen

PR #306, opened 2016-03-14
Remove count guards around removeAnnotation and removeAllAnnotations

tmcw:
cc @bsudekum - it doesn't appear that these guards prevent any error
or give any appreciable performance boost. I believe they should be
removed.


bsudekum:
:+1:

PR #312, opened 2016-03-22
Add ios offline support

bsudekum:
Adds support for ios offline.

Currently the api looks like: 

``` js
 this.addPackForRegion(mapRef, {
  name: 'test',
  bounds: [0, 0, 0, 0],
  minZoomLevel: 0,
  maxZoomLevel: 0,
  style: this.mapStyles.emerald
});
```

`bounds` is a tricky one to make future-proof-able because new `regions` will come out down the line ie offline along a line. I might add a `type` to this.

still todo:
- [x] fixup `getPacksWithCompletionHandler` 

/cc @1ec5 for :eyes:

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -55,6 +55,9 @@ var MapMixins = {
   getDirection(mapRef, callback) {
     NativeModules.MapboxGLManager.getDirection(React.findNodeHandle(this.refs[mapRef]), callback);
   },
+  getBounds(mapRef, callback) {
+    NativeModules.MapboxGLManager.getBounds(React.findNodeHandle(this.refs[mapRef]), callback);
+  },
   mapStyles: NativeModules.MapboxGLManager.mapStyles,
   userTrackingMode: NativeModules.MapboxGLManager.userTrackingMode,
   userLocationVerticalAlignment: NativeModules.MapboxGLManager.userLocationVerticalAlignment,
--- a/index.ios.js
+++ b/index.ios.js
@@ -4,6 +4,12 @@ var React = require('react-native');
 var { NativeModules, requireNativeComponent } = React;
 
 var MapMixins = {
+  addPackForRegion(mapRef, options) {
+    NativeModules.MapboxGLManager.addPackForRegion(React.findNodeHandle(this.refs[mapRef]), options);
+  },
+  getPacksWithCompletionHandler(mapRef, callback) {
+    NativeModules.MapboxGLManager.getPacksWithCompletionHandler(React.findNodeHandle(this.refs[mapRef]), callback);
+  },
   setDirectionAnimated(mapRef, heading) {
     NativeModules.MapboxGLManager.setDirectionAnimated(React.findNodeHandle(this.refs[mapRef]), heading);
   },
@@ -48,7 +54,8 @@ var MapMixins = {
   },
   mapStyles: NativeModules.MapboxGLManager.mapStyles,
   userTrackingMode: NativeModules.MapboxGLManager.userTrackingMode,
-  userLocationVerticalAlignment: NativeModules.MapboxGLManager.userLocationVerticalAlignment
+  userLocationVerticalAlignment: NativeModules.MapboxGLManager.userLocationVerticalAlignment,
+  unknownResourceCount: NativeModules.MapboxGLManager.unknownResourceCount
 };
 
 var MapView = React.createClass({
@@ -81,6 +88,15 @@ var MapView = React.createClass({
   },
   _onLocateUserFailed(event: Event) {
     if (this.props.onLocateUserFailed) this.props.onLocateUserFailed(event.nativeEvent.src);
+  },
+  _onOfflineProgressDidChange(event: Event) {
+    if (this.props.onOfflineProgressDidChange) this.props.onOfflineProgressDidChange(event.nativeEvent.src);
+  },
+  _onOfflineMaxAllowedMapboxTiles(event: Event) {
+    if (this.props.onOfflineMaxAllowedMapboxTiles) this.props.onOfflineMaxAllowedMapboxTiles(event.nativeEvent.src);
+  },
+  _onOfflineDidRecieveError(event: Event) {
+    if (this.props.onOfflineDidRecieveError) this.props.onOfflineDidRecieveError(event.nativeEvent.src);
   },
   propTypes: {
     showsUserLocation: React.PropTypes.bool,
@@ -134,7 +150,10 @@ var MapView = React.createClass({
     onLocateUserFailed: React.PropTypes.func,
     onLongPress: React.PropTypes.func,
     contentInset: React.PropTypes.array,
-    userLocationVerticalAlignment: React.PropTypes.number
+    userLocationVerticalAlignment: React.PropTypes.number,
+    onOfflineProgressDidChange: React.PropTypes.func,
+    onOfflineMaxAllowedMapboxTiles: React.PropTypes.func,
+    onOfflineDidRecieveError: React.PropTypes.func
   },
   getDefaultProps() {
     return {
@@ -167,7 +186,10 @@ var MapView = React.createClass({
         onLongPress={this._onLongPress}
         onFinishLoadingMap={this._onFinishLoadingMap}
         onStartLoadingMap={this._onStartLoadingMap}
-        onLocateUserFailed={this._onLocateUserFailed} />
+        onLocateUserFailed={this._onLocateUserFailed}
+        onOfflineProgressDidChange={this._onOfflineProgressDidChange}
+        onOfflineMaxAllowedMapboxTiles={this._onOfflineMaxAllowedMapboxTiles}
+        onOfflineDidRecieveError={this._onOfflineDidRecieveError} />
     );
   }
 });
--- a/index.ios.js
+++ b/index.ios.js
@@ -7,8 +7,8 @@ var MapMixins = {
   addPackForRegion(mapRef, options) {
     NativeModules.MapboxGLManager.addPackForRegion(React.findNodeHandle(this.refs[mapRef]), options);
   },
-  getPacksWithCompletionHandler(mapRef, callback) {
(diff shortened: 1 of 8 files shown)


ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -106,6 +111,25 @@ - (void)createMap
     [self layoutSubviews];
 }
 
+-(void)createOfflinePack:(MGLCoordinateBounds)bounds style:(NSURL*)style fromZoomLevel:(double)fromZoomLevel toZoomLevel:(double)toZoomLevel name:(NSString*)name
+{
+    id <MGLOfflineRegion> region = [[MGLTilePyramidOfflineRegion alloc] initWithStyleURL:style bounds:bounds fromZoomLevel:fromZoomLevel toZoomLevel:toZoomLevel];
+    
+    NSDictionary *userInfo = @{ @"name": name };
```
1ec5: This is pretty limiting. Clients should be able to store more than just a name with each pack, for instance a date, icon, or some other application-specific data. I think the name parameter should be replaced with an NSDictionary-typed userInfo parameter.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -504,6 +528,30 @@ - (void)mapViewWillStartLoadingMap:(MGLMapView *)mapView
...
+                                     }
+                             };
+    
+    [_eventDispatcher sendInputEventWithName:@"onProgressDidChange" body:event];
+}
+
+- (void)offlinePack:(MGLOfflinePack *)pack didReceiveMaximumAllowedMapboxTiles:(uint64_t)maximumCount {
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"maximumCount": @true
```
1ec5: Don't use true in Objective-C; use YES.

But why is this method returning a Boolean anyways? A numeric maximum count should be sent to the client so it knows what the tile limit is.

1ec5:
The delegate is missing the -offlinePack:didRecieveError: method. That method is technically optional, but I think you'd want clients to be able to react to error conditions. Note that some of the errors that trigger this delegate callback are recoverable, so you should provide enough information in the resulting event for clients to distinguish between different error conditions. (You could send along the error code in the event and also expose constants for each of the possible error codes.)

tmcw:
Should this method return an event emitter or call a callback instead of these callbacks being map-object-wide?

bsudekum:
@tmcw which method are you looking at?

tmcw:
`this.addPackForRegion` triggers a download and then calls `onOfflineProgressDidChange or`onOfflineMaxAllowedMapboxTiles`. Instead, I'm saying that it should be more like

``` js
 this.addPackForRegion(mapRef, {
  name: 'test',
  bounds: [0, 0, 0, 0],
  minZoomLevel: 0,
  maxZoomLevel: 0,
  style: this.mapStyles.emerald
}).on('error', function() {}).on('progress', function {});
```

or such, because that'd be how we'd write it if it was a pure JS implementation.

bsudekum:
@tmcw you can listen on the following event emitters with this pr:
- `onOfflineProgressDidChange`
- `onOfflineMaxAllowedMapboxTiles`
- `onOfflineDidRecieveError` (Forgot one step in hooking this up now that im looking throw the code)

These are normal props (functions) that you can add to your mapview. Yes, `.on('error', function() {}).on('progress', function {});` is one way to skin the cat, but all other events are props on the mapview currently. I'd like to keep it this way and preserve this pattern.

tmcw:
I feel like this is a pattern from obj-c and it doesn't work that well in JavaScript-land. Returning an event emitter for a single download and calling callbacks from the map object for all downloads are different things from the developers perspective - the latter lets you download multiple offline packs and have to match up which progress event goes to which.

That said, I totally see how we should land this asap and not get caught up in semantic quibbles. Long-term though I think we should aim to make this API more friendly to JS-land and similar to existing React patterns.

index.ios.js
```diff
@@ -4,6 +4,15 @@ var React = require('react-native');
 var { NativeModules, requireNativeComponent } = React;
 
 var MapMixins = {
+  addPackForRegion(mapRef, options) {
+    NativeModules.MapboxGLManager.addPackForRegion(React.findNodeHandle(this.refs[mapRef]), options);
+  },
+  getPacksWithCompletionHandler(mapRef, callback) {
```
1ec5: There's no longer any reason to make this asynchronous. The underlying property is completely synchronous now.

1ec5:
I agree with @tmcw: especially given the presence of the error notification, a more JavaScripty design would involve Promises somehow. Implementing that here could be pretty straightforward, but you'd need to register a block-based KVO change observer right after resuming a pack, rather than relying on notifications.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -104,12 +109,34 @@ - (void)createMap
...
     [self addSubview:_map];
     [self layoutSubviews];
 }
 
+-(void)createOfflinePack:(MGLCoordinateBounds)bounds style:(NSURL*)style fromZoomLevel:(double)fromZoomLevel toZoomLevel:(double)toZoomLevel name:(NSString*)name type:(NSString*)type
+{
+    
+    id <MGLOfflineRegion> region = [[MGLTilePyramidOfflineRegion alloc] initWithStyleURL:style bounds:bounds fromZoomLevel:fromZoomLevel toZoomLevel:toZoomLevel];
+    
+    NSDictionary *userInfo = @{ @"name": name };
```
1ec5: I still think this is too limited. Why not allow other dictionary entries to be set too?

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -502,6 +529,54 @@ - (void)mapViewWillStartLoadingMap:(MGLMapView *)mapView
...
+
+- (void)offlinePackDidReceiveMaximumAllowedMapboxTiles:(NSNotification *)notification {
+    MGLOfflinePack *pack = notification.object;
+    NSDictionary *userInfo = [NSKeyedUnarchiver unarchiveObjectWithData:pack.context];
+    uint64_t maximumCount = [notification.userInfo[MGLOfflinePackMaximumCountUserInfoKey] unsignedLongLongValue];
+    
+    NSDictionary *event = @{ @"target": self.reactTag,
+                             @"src": @{
+                                     @"name": userInfo[@"name"],
+                                     @"maximumCount": @(maximumCount)
```
1ec5: In JavaScript, I think you'd typically name this `maxTiles`.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -165,6 +170,82 @@ - (NSDictionary *)constantsToExport
...
+}
+
+RCT_EXPORT_METHOD(getPacksWithCompletionHandler:(nonnull NSNumber *)reactTag
+                  findEvents:(RCTResponseSenderBlock)callback)
+{
+    [_bridge.uiManager addUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber *, RCTMapboxGL *> *viewRegistry) {
+        RCTMapboxGL *mapView = viewRegistry[reactTag];
+        if ([mapView isKindOfClass:[RCTMapboxGL class]]) {
+            RCTMapboxGL *mapView = viewRegistry[reactTag];
+            NSMutableArray* callbackArray = [NSMutableArray new];
```
1ec5, line 234: You could significantly simplify this function by taking all this code out of the block and returning this array instead of calling a completion handler.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -165,6 +170,82 @@ - (NSDictionary *)constantsToExport
...
+            
+            MGLOfflinePack *packs = [MGLOfflineStorage sharedOfflineStorage].packs;
+            BOOL hasDeletedAPack = NO;
+            
+            for (MGLOfflinePack *pack in packs) {
+                NSDictionary *userInfo = [NSKeyedUnarchiver unarchiveObjectWithData:pack.context];
+                if (userInfo[@"name"] == packName) {
+                    [[MGLOfflineStorage sharedOfflineStorage] removePack:pack withCompletionHandler:^(NSError * _Nullable error) {
+                        if (error != nil) {
+                            return RCTLogError(@"Error: %@", error.localizedFailureReason);
```
1ec5: Remove the return keywords in this block. They just cause the block to return, but the for loop containing the block will continue looping, which is probably not what you intended.

ios/RCTMapboxGL/RCTMapboxGL.m
bsudekum: Although the internal api allows for more flexibility, these bindings need to be a little more strict. Having it strict allows for the remove pack logic to be a little more [straight forward](https://github.com/mapbox/react-native-mapbox-gl/pull/312/files#diff-d0ace68cb3e436e36c7bb0bca305815cR216). If it was an open ended dictionary that the dev decided upon, I think it would be more difficult to know which pack to remove.

1ec5: You could still identify packs by a `name` key, or better yet by an `id` key, but it should definitely be possible to add information other than an ID alongside the ID.

ios/RCTMapboxGL/RCTMapboxGLManager.m
1ec5, line 234: Per chat, I was mistaken: this block is used for asynchronously getting access to MGLMapView.

index.ios.js
1ec5: @bsudekum explained to me that this unfortunately still has to be asynchronous because _everything_ is asynchronous in this component, even where there’s no technical reason for asynchronicity. You may still hear me grumble loudly, but it’s a design limitation that’s apparently out of this component’s control.

PR #321, opened 2016-04-06
Bring back onTap

magus:
Noticed this was removed, probably because it didn't originally work well with annotations and swallowed the tap event? The plumbing here should fix that and safely trigger both.

It's great for tracking the users last touch. I'm going to use it to draw custom callouts with React native views where the user last touched (including over annotations).


bsudekum:
Thanks!

PR #331, opened 2016-04-19
Allow empty array of annotations

jayfunk:
If you provide to the MapView component an empty array of annotations an error is logged. This will prevent the map view from being updated. In my current project I filter the annotations based on certain criteria and it is potential have zero annotations match the filter. In this case I expect the map view to have all annotations removed. This works in iOS.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -31,7 +31,7 @@ repositories {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.19.+'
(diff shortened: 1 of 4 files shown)


bsudekum:
@jayfunk It looks like you branched off my v4.0.0 branch which is not ready. Can you rebase this so it does not bring in my changes in https://github.com/mapbox/react-native-mapbox-gl/pull/331/commits/059e418e1eae5bc4709078cddd1ace6a61bc8cd5?

PR #333, opened 2016-04-19
Removed check for empty array of annotations

jayfunk:
If you provide to the MapView component an empty array of annotations an error is logged. This will prevent the map view from being updated. In my current project I filter the annotations based on certain criteria and it is potential have zero annotations match the filter. In this case I expect the map view to have all annotations removed. This works in iOS.


jayfunk:
Sorry I created another pull request. I was frustrated with the rebase. :)

bsudekum:
👍

PR #337, opened 2016-04-27
Add height and width support to annotationImage (android)

cl3m:
The current android implementation does not support height and width like on ios.


bsudekum:
👍 thanks @cl3m!

PR #343, opened 2016-05-04
add required empty metadata object to addPackForRegion.

timfpark:
Without the empty metadata object, clicking "Create offline pack" results in an red box error that "metadata is required".


bsudekum:
👍 thanks!

Issue #344, opened 2016-05-05
Support for react-native 0.25 and beyond


urbancvek:
Hey. react-native 0.25.0 requires to import React from 'react' instead from 'react-native'. This means that you cannot use this project in any version beyond 0.25. Would you accept a PR if I submit one?

I have it working for myself but would like to push solution here. 

It just requires to import React from 'react' and then import ReactNative from 'react-native'. 
Then you have to decouple PropTypes, Component from React and other stuff like View, Text, StyleSheet and anything else from ReactNative.

PR #345, opened 2016-05-06
Added support for react-native 0.25 and beyond

urbancvek:
Hey I opened this issue #344 yesterday about adding support for react-native 0.25 and up. I updated only the iOS part so you can take a look at what I did and tell me if you like the changes. I would of course update the android part too before the PR would be merged in.

I also did some refactoring of the code. Extract PropTypes, findNodeHandle (I did everything in separate commits so you can easily take a look at what I changed)

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var React = require('react');
+var { PropTypes } = React;
+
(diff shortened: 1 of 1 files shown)


bsudekum:
Awesome, thanks @urbancvek. Tested, works great.

PR #346, opened 2016-05-06
Added Android support for react-native 0.25 and up

urbancvek:
Did the same thing for Android as before for iOS.


bsudekum:
Awesome, thanks again @urbancvek!

PR #349, opened 2016-05-08
setCenterCoordinateAnimated returns a promise

### LINKED ISSUE #348: PR: Methods for Modifying the Map State should return a Promise
Locally I've made some changes to allow the `setCenterCoordinateAnimated` method to return a promise which resolves after the action is completed using the `completionHandler` argument to resolve the promise. This is useful I've found for doing things like centering on an annotation after opening it. I'd be willing to do a PR to add this to all the methods that have a `completionHandler` but don't want to invest if it won't be merged. Has anyone considered this or think it would be a useful addition to add?

``` objc
-(void)setCenterCoordinateAnimated:(CLLocationCoordinate2D)coordinates resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject
{
    [_map setCenterCoordinate:coordinates zoomLevel:_map.zoomLevel direction:_map.direction animated:YES completionHandler:^{
        resolve(@"DONE");
    }];
}
```

magus:
close #348 
- Hook into completionHandler to resolve promise

See issue for explanation. If we like this pattern I wouldn't mind extending it to the other methods as well where there is a completionHandler we can leverage. It shouldn't break any existing usage, just returns a promise instead of the current `undefined`.

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -25,7 +25,7 @@ var MapMixins = {
     MapboxGLManager.setZoomLevelAnimated(findNodeHandle(this.refs[mapRef]), zoomLevel);
   },
   setCenterCoordinateAnimated(mapRef, latitude, longitude) {
-    MapboxGLManager.setCenterCoordinateAnimated(findNodeHandle(this.refs[mapRef]), latitude, longitude);
+    return MapboxGLManager.setCenterCoordinateAnimated(findNodeHandle(this.refs[mapRef]), latitude, longitude);
   },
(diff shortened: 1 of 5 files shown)


magus:
Example usage

``` js
this.setCenterCoordinateAnimated(mapRef, annotation.latitude, annotation.longitude).then(() => {
  this.setState({ overlay: annotation });
});
```

bsudekum:
This is awesome, thanks @magus

PR #352, opened 2016-05-12
Added missing source_files directive

ghost:
Hey there,

The CocoaPod installation is currently not working; - only the MapBox framework is being linked (the RTC bridge module hasn't been included).


bsudekum:
Ah, good catch. Thanks!

Issue #357, opened 2016-06-02
mapbox - removepack feature is not working correctly.


grossingdev:
NSString comparison must be progressed using isEqualToString, not using == operator.

in removePack function of RCTMapboxGL.m

timfpark:
I see this in my app as well: submitted https://github.com/mapbox/react-native-mapbox-gl/pull/359 to fix.

PR #359, opened 2016-06-09
Fix string equality check in removePack

timfpark:
Fix string equality per https://github.com/mapbox/react-native-mapbox-gl/issues/357


bsudekum:
👍  thanks!

Issue #361, opened 2016-06-12
Mixins and this.refs are deprecated


dapetcu21:
React is in the process of deprecating this.refs and mixins. The correct way of implementing imperative methods is by attaching the methods directly to the component class, so the user can do 
`<MapView ref={map => { this.mapRef = map; }} />`, then the simpler `this.mapRef.addAnnotations(annotations)`.

Mixins are unusable, for example, if you use the new ES6 `class MyComponents extends Component {...}` syntax.

This transition to this API can be implemented in a backwards compatible fashion. I am willing to submit a PR for this if you agree that this is what should be done.

urbancvek:
I am not a maintainer of this project but I also agree with you and I thought about it myself. 
All map methods should than be rewritten without first argument which is now MAP_REF.

If you decide to do it I would be willing to help. 😄

dapetcu21:
I'll do it. It shouldn't take me more than a few hours. I'll actually make a bigger PR (#365) with some more API improvements. Any help on an Android port for these changes will be appreciated.

dapetcu21:
Closing this after fixing it in #365

PR #365, opened 2016-06-17
Re-design API more into the React sprit. Unify Android and iOS APIs

dapetcu21:
The API in its current state has several issues in terms of abiding by React principles and, frankly, would need a major redesign. This PR aims to solve that, albeit at the expense of backwards compatibility.

Additionally, since I will use the library for an app I'm working on which needs to be cross platform, I'll unify the Android and iOS APIs.

Firstly, as explained in #361, the mixins API needs to go away and has to be replaced with class methods on the component.

A few calls (like the offline pack API and setting the API key) need to become independent of the component (static methods). #362

Also, there is a number of imperative methods that could be re-implemented as props like `setUserTrackingMode` and the `annotations` family.

Props that only set initial states of the component need to be renamed with `initial` prepended, as recommended in [the React docs](https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html)

The component must also respond gracefully to props changes (it's not acceptable to reset the whole state of the component and break user interaction every time any prop is changed). This involves getting rid of `-(void)updateMap`

To do list:
- [x] Replace the mixin API with class methods
- [x] Separate `accessToken` from component as a call to a static method
- [x] Separate the offline methods as static methods
- [x] Make the native component respond gracefully to individual props changes
- [x] Implement `onChangeUserTrackingMode` event
- [x] Implement the metrics API
- [x] Prepend `initial` to `zoomLevel`, `centerCoordinate`, `direction`
- [x] Remove methods mutating `annotations` in favor of changing the props.
- [x] Diff `annotations` before sending only the changes through the bridge.
- [x] Rename component from `Mapbox` to `MapView`
- [x] Refactor events not to use `customDirectEventTypes`
- [x] Rename all `blahBlahAnimated` methods as `blahBlah` with an `animated` argument
- [x] Document the whole thing
- [x] Throttle events that were starving the JS run loop.
- [x] Unify Android and iOS APIs
- [x] Backport events to Android
- [x] Backport annotations to Android
- [x] Implement offline packs functionality for Android


dapetcu21:
Oh god...
screen shot 2016-06-20 at 16 47 41

dapetcu21:
And with this, I think the PR is good to go. There were a lot of changes. Take a look over the docs to get a clear big picture.

dapetcu21:
@bsudekum Can you take a look?

dapetcu21:
I started work on the Android port.

dapetcu21:
And with this, the Android port is done. Features should be on-par with iOS and most bugs have been addressed. I think the library is now much closer to a production ready state. Are there any maintainers around to take a look?

API.md
```diff
@@ -0,0 +1,440 @@
...
+| `onLongPress` | `{latitude: 0, longitude: 0, screenCoordX: 0, screenCoordX: 0}` | Fired when the user taps and holds screen for 1 second.
+| `onStartLoadingMap` | `undefined` | Fired once the map begins loading the style. |
+| `onFinishLoadingMap` | `undefined` | Fired once the map has loaded the style. |
+
+## Methods
+
+You first need to get a ref to your `MapView` component:
+
+```jsx
+<MapView ref={map => { this._map = map; }} />
```
bsudekum, line 72: Why is it necessary for ref to be a function? Could it just be 

```
ref=`map`
```

and then

```
this.refs.map.setDirection()
```

API.md
```diff
@@ -0,0 +1,440 @@
...
+The transition is animated unless you pass `animated` as `false`.
+
+## Styles
+
+#### Default styles
+
+Mapbox GL ships with 6 included styles:
+
+* `Mapbox.mapStyles.streets`
+* `Mapbox.mapStyles.emerald`
```
bsudekum: Emerald has been removed in v9 styles.

API.md
dapetcu21, line 72: From here: https://facebook.github.io/react/docs/more-about-refs.html
"Although string refs are not deprecated, they are considered legacy, and will likely be deprecated at some point in the future. Callback refs are preferred."

android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
```diff
@@ -0,0 +1,225 @@
...
+
+    static Map<String, Icon> iconCache = new HashMap();
+
+    static Icon iconFromPathAndSize(Context context, String path, int width, int height) throws IOException {
+        String cacheKey = path + "||" + width + "||" + height;
+        Icon icon = iconCache.get(cacheKey);
+        if (icon != null) { return icon; }
+
+        Drawable drawable;
+        if (path.startsWith("image!")) {
```
bsudekum: I wonder if we should depreciate `image!` with 5.0.0. `image!` has not been part of rn for a long time now. Perhaps we should look into passing a required image through.

android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
```diff
@@ -0,0 +1,225 @@
...
+            String subtitle = annotation.getString("subtitle");
+            marker.snippet(subtitle);
+        }
+
+        if (annotation.hasKey("annotationImage")) {
+            ReadableMap annotationImage = annotation.getMap("annotationImage");
+            String annotationURL = annotationImage.getString("url");
+            try {
+                int width = -1;
+                int height = -1;
```
bsudekum, line 156: Curious, why -1 as defaults?

API.md
dapetcu21: Nice to know. I just shipped an app with emerald. I can add a deprecation warning to it.

android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
dapetcu21: Agreed, though I have no idea how to do that. It needs to be researched.

dapetcu21: I say leave it like this for now, then deprecate it later if we find a better solution.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLModule.java
```diff
@@ -59,87 +119,414 @@ public String getName() {
...
+
+        @Override
+        public void onError(OfflineRegionError error) {
+
+        }
+
+        @Override
+        public void mapboxTileCountLimitExceeded(long limit) {
+
+        }
```
bsudekum: Why are these empty?

android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
dapetcu21, line 156: Just a small convention to say "intrinsic size of the drawable". See L117 above.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
```diff
@@ -0,0 +1,656 @@
...
+    @Override
+    public void onHostPause() {
+        _paused = true;
+        _mapView.onPause();
+    }
+
+    @Override
+    public void onHostDestroy() {
+        onDrop();
+    }
```
bsudekum, line 124: LifeCycle 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 !

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
```diff
@@ -0,0 +1,656 @@
...
+            event.putMap("src", src);
+            emitEvent("onLocateUserFailed", event);
+            return;
+        }
+
+        src.putDouble("latitude", location.getLatitude());
+        src.putDouble("longitude", location.getLongitude());
+        src.putDouble("magneticHeading", location.getBearing());
+        src.putDouble("trueHeading", location.getBearing());
+        src.putDouble("headingAccuracy", 10.0);
```
bsudekum: Should this be hard coded to 10.0?

example.js
```diff
@@ -0,0 +1,307 @@
...
+  };
+  onLongPress = (location) => {
+    console.log('onLongPress', location);
+  };
+  onTap = (location) => {
+    console.log('onTap', location);
+  };
+  onChangeUserTrackingMode = (userTrackingMode) => {
+    this.setState({ userTrackingMode });
+    console.log('onChangeUserTrackingMode', userTrackingMode);
```
bsudekum, line 93: Might be better to re-write as 

``` js
this.setState({ userTrackingMode }, () => {
  console.log('onChangeUserTrackingMode', userTrackingMode);
});
```

So you know that state has been set.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLModule.java
dapetcu21: After removing a region, mapbox still calls methods on the observer, so I just swapped the OfflineRegionProgressObserver with a dummy one. Later I found that this doesn't work as the observer change doesn't go into effect, so I added an `invalidated` flag in OfflineRegionProgressObserver. Therefore, I guess this class can go.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
dapetcu21: Not really, but the Android SDK didn't give me `headingAccuracy`. We could remove it and live with the idea that iOS and Android would have different signatures for this event or we could set it to -1 or some other value, mention it in the docs and keep consistency across platforms.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -205,396 +185,356 @@ - (UIColor *)mapView:(MGLMapView *)mapView fillColorForPolygonAnnotation:(RCTMGL
...
+    if (!_onRegionWillChange) { return; }
 
-    return annotationImage;
+    CLLocationCoordinate2D region = _map.centerCoordinate;
+    _onRegionWillChange(@{ @"target": self.reactTag,
+                           @"src": @{ @"latitude": @(region.latitude),
+                                      @"longitude": @(region.longitude),
+                                      @"zoomLevel": @(_map.zoomLevel),
+                                      @"direction": @(_map.direction),
+                                      @"pitch": @(0),
```
bsudekum: Should prob be the actual pitch, not zero.

example.js
dapetcu21, line 93: I'm not sure how many people are aware that `this.setState` receives a callback. For the clarity of the example code, I think it's better the way it is. Or maybe with the `console.log` before `setState`.

bsudekum:
This is really awesome @dapetcu21, thanks for your time and hard work. I just gave it a once over and things look really great. I'd like to pull in @1ec5 and @bleege who both work on mapbox-gl-native on iOS and Android respectfully. If you guys could check out the obj-c and Android code that would be appreciated. Gunna head out and grill because 🇺🇸 🎆  but can take a look at this again tomorrow.

Todo:
- [ ] Make @dapetcu21 a collaborator on this project
- [ ] @bleege @1ec5 give feedback
- [ ] @bsudekum pull and test code
- [ ] fix with feedback
- [ ] @dapetcu21 merges

ios/RCTMapboxGL/RCTMapboxGL.m
dapetcu21: Couldn't find the getter for that in the iOS SDK. Camera is implemented really weirdly in Mapbox for iOS.

bsudekum: Ok, let me see what I can do. We should either remove it, or implement it.

dapetcu21:
Thanks. Have fun at the grill 🍖

API.md
dapetcu21: Added the warning

android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
dapetcu21: Actually, we could change it from `url` to `source: { uri: 'myImage' }` to imitate the `<Image />` API, then see in the future how can we integrate it with `require()` without changing the API.

ios/RCTMapboxGL/RCTMapboxGL.m
1ec5: Use [`_map.camera.pitch`](https://www.mapbox.com/ios-sdk/api/3.3.0-beta.3/Classes/MGLMapCamera.html#/c:objc%28cs%29MGLMapCamera%28py%29pitch). The API is consistent with MapKit, which is what we’re going for in the Mapbox iOS SDK.

dapetcu21: Ok. Thanks. Will change. Never used MapKit. Can we create cameras from zoomLevels instead of altitudes in the future, though? Or maybe add pitch to `setCenterCoordinate:zoomLevel:direction:animated:completionHandler:` Otherwise, I can't see an easy way to ease to a known position, heading, zoomLevel and pitch (to match the Android API). Am I missing something?

API.md
```diff
@@ -0,0 +1,440 @@
...
+Then call methods as `this._map.methodName()`.
+
+---
+
+```javascript
+this._map.setDirection(direction, animated = true, callback);
+this._map.setZoomLevel(zoomLevel, animated = true, callback);
+this._map.setCenterCoordinate(latitude, longitude, animated = true, callback);
+this._map.setCenterCoordinateZoomLevel(latitude, longitude, zoomLevel, animated = true, callback);
+this._map.setCenterCoordinateZoomLevelPitch(latitude, longitude, zoomLevel, pitch, animated = true, callback); // Android only
```
1ec5: The equivalent on iOS would be to set the map camera. On either platform, when the map is tilted, the map’s “zoom level” doesn’t actually correspond to the zoom level of the tiles that are visible on the map. So MapKit and the Mapbox iOS SDK let you [create a camera](https://www.mapbox.com/ios-sdk/api/3.3.0-beta.3/Classes/MGLMapCamera.html) based on the altitude above the ground or the distance from the focal point.

API.md
```diff
@@ -0,0 +1,440 @@
...
+this._map.setPitch(pitch, animated = true, callback); // Android only
```
1ec5: This could be implemented on iOS as well by setting MGLMapView’s `camera.pitch` property.

API.md
1ec5: Ah, I see that you’ve addressed this down below. 👍

API.md
```diff
@@ -0,0 +1,440 @@
...
+
+```javascript
+this._map.setCamera(latitude, longitude, fromDistance, pitch, direction, animationDuration = 0.3); // iOS Only
+```
+
+This method only works on iOS.
+
+Sets the map pitched at an angle (`pitch`, measured in
+degrees, 0 is from straight above), looking towards `latitude` and `longitude`
+from `fromDistance` meters away, pointing towards heading `direction`.
```
1ec5: It’s also possible to create an MGLMapCamera with a viewing distance instead of an altitude. Depending on your use case, you might want one or the other without having to do the trig yourself.

API.md
dapetcu21: Yeah, but the Android SDK doesn't have such function. So what do I expose to the user? Setting altitude or setting zoomLevel? Android has zoomLevel, but no altitude. iOS has altitude but no zoomLevel.

So how do I solve the following scenario with a cross platform API?
"The user wants to change both pitch and zoomLevel at the same time."

API.md
dapetcu21: True, but there is no alternative for `_map.setCenterCoordinateZoomLevelPitch`, since that sets both pitch (which is not available with `setCenterCoordinate:blahblahblah`) and zoomLevel, which is not available with `setCamera:`.

1ec5:
Some big changes are coming in [iOS SDK v3.3.0](https://www.mapbox.com/ios-sdk/api/3.3.0-beta.3/), currently in beta. Any thoughts as to how this library should handle annotation views (also in Android SDK v4.1.0 as view-based markers) and feature querying?

API.md
1ec5: Opened mapbox/mapbox-gl-native#5583 to track a convenience API for this functionality.

dapetcu21:
@1ec5 I skimmed the API docs a bit. I suppose feature querying should be exposed as a method on the component, like the rest.  I would stay away from annotation views for now. I don't know how/if you can render react components from within native components. It's certainly not documented, anyhow.

ios/RCTMapboxGL/RCTMapboxGL.m
1ec5: Tracking in mapbox/mapbox-gl-native#5583. I don’t think it’s best to mix 2D and 3D metaphors, especially if we ever start to vary zoom levels based on viewing distance or introduce an additional `roll` degree of freedom. 😉 But we can provide convenience methods to bridge the two models.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
bleege: The closest that the stock Android SDK's `Location` object provides is [`getAccuracy()`](https://developer.android.com/reference/android/location/Location.html#getAccuracy%28%29).

bleege:
@dapetcu21 Great job bringing this together!  Before merging this can you confirm that it's been tested against the Android 4.1.0 Final SDK that was released last week?

dapetcu21:
Well, I started working on it kinda exactly when 4.1.0 came out, so I guess it works.

android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
dapetcu21: Yeah. We could expose accurracy on Android separately, but I'd rather not serve meters as degrees.

1ec5: `getAccuracy()` is the accuracy of the location, not the heading:

> This accuracy estimation is only concerned with horizontal accuracy, and does not indicate the accuracy of bearing, velocity or altitude if those are included in this Location.

The analogue to that on iOS would be [`CLLocation.horizontalAccuracy`](https://developer.apple.com/reference/corelocation/cllocation/1423599-horizontalaccuracy).

dapetcu21:
@bsudekum @bleege @1ec5 I integrated some of the feedback in the past few commits.

ios/RCTMapboxGL/RCTMapboxGLConversions.m
```diff
@@ -54,48 +60,32 @@
...
-            point.annotationImageURL = annotationImageURL;
-            point.annotationImageSize = annotationImageSize;
+        point = [[RCTMGLAnnotation alloc] initWithLocation:CLLocationCoordinate2DMake(lat, lng) title:title subtitle:subtitle id:id];
+    }
+    
+    if ([annotationObject valueForKey:@"annotationImage"]) {
+        NSDictionary *annotationImage = [annotationObject valueForKey:@"annotationImage"];
+        NSDictionary *imageSource = (NSDictionary*)annotationImage[@"source"];
+        CGFloat height = (CGFloat)[[annotationImage valueForKey:@"height"] floatValue];
+        CGFloat width = (CGFloat)[[annotationImage valueForKey:@"width"] floatValue];
```
1ec5: `CGFloat` is actually a typedef for `double`, so you can avoid the cast by calling `-doubleValue` instead of `-floatValue`.

1ec5:
This is great @dapetcu21! No concerns from me on the iOS side; your changes are a vast improvement.

ios/RCTMapboxGL/RCTMapboxGLConversions.m
dapetcu21: Oh. That's from the old code. Never knew that `CGFloat` is actually a `double`. Cool.

1ec5: I take that back slightly: it’s typedef’d to a `float` on 32-bit devices and a `double` on 64-bit devices.

dapetcu21: Actually, using `doubleValue` still stands.

tlvenn:
Any chance to merge this soon ?

package.json
```diff
@@ -24,7 +24,7 @@
   "scripts": {
     "preinstall": "./scripts/download-mapbox-gl-native-ios.sh 3.2.0",
```
bsudekum: Can we change this to v3.3.1?

dapetcu21: I guess so. Didn't test yet. Let's merge though.

afischer:
This is awesome stuff—excited to see this merged.

JohnyDays:
Awesome, great work @dapetcu21 and everyone! 🎈

package.json
bsudekum: @dapetcu21 I'd like to get both android and ios on the latest version. We just released a big [bug fix release](http://status.mapbox.com/incidents/59f09r86vhyt) and I'd like to make sure we're on this. Theoretically, there should be no breaking changes since this branch is already on 3.x.

dapetcu21: Ok. Then I'll check it out when I get home.

dapetcu21: Done. I upgraded both the iOS and Android SDKs. They seem to work fine on a first glance. Let's merge this PR, please :D

dapetcu21:
@bsudekum Let's merge this soon, as my fork seems to be diverging from master. Also, let's release it to npm after the merge.

dapetcu21:
I merged master into this to fix the conflicts. Should be ready for merging.

tmcw:
👍 🚢 

Thanks @dapetcu21 for the hard work on this PR! Merging now

dapetcu21:
Btw, I noticed there's a large amount of issues and PRs that are either fixed or superseeded by this merge. They should be cleaned up.

tmcw:
👍 @dapetcu21 I've added you as a contributor to this project so you have the power to push commits to this repo and close issues.

dapetcu21:
@tmcw Thanks a lot. I appreciate it.

dapetcu21:
I promise I won't push to master, haha

PR #367, opened 2016-06-21
fix: defaultProps method name for android

yutin1987:
ref.
https://facebook.github.io/react/docs/reusable-components.html#default-prop-values


tmcw:
👍 Thanks!

Issue #350, opened 2016-05-08
Build failed due to 'Mapbox/Mapbox.h' file not found


ds3783:
After project configuration changing, which I followed installation guide, the project could't build. 
Error message as follows:

~/Projects/my/AwsomeApp/node_modules/react-native-mapbox-gl/ios/RCTMapboxGL/RCTMapboxGL.h:9:9: 'Mapbox/Mapbox.h' file not found

~/Projects/my/AwsomeApp/node_modules/react-native-mapbox-gl/ios/RCTMapboxGL/RCTMapboxGL.m:9:9: In file included from ~/Projects/my/AwsomeApp/node_modules/react-native-mapbox-gl/ios/RCTMapboxGL/RCTMapboxGL.m:9:

jayfunk:
What do you mean by configuration changing? Is the mapbox.h file actually
there? Did you update mapbox?

ds3783:
I followed installation manual,but my project couldn't compile

https://github.com/mapbox/react-native-mapbox-gl/blob/master/ios/install.md

my mapbox sdk is 3.2.0

yutin1987:
+1
I follow this step:
https://www.mapbox.com/ios-sdk/#manual

yutin1987:
This issue occurs when npm install is failed (run download-mapbox-gl-native-ios.sh is failed).
ref. #366

yutin1987:
You can try reinstall

```
npm uninstall react-native-mapbox-gl --save
npm install react-native-mapbox-gl --save
```

dapetcu21:
True, the installation documentation needs to be updated to include adding Mapbox to embedded binaries.

digitaldavenyc:
+1 for this issue

1ec5:
Via #402 and #417, the CocoaPods installation method doesn’t automatically link Mapbox.framework correctly either.

ghost:
Jup, I have the same problem, any fix??

ghost:
DONE!

Download the Framework from this Link.

http://mapbox.s3.amazonaws.com/mapbox-gl-native/ios/builds/mapbox-ios-sdk-3.3.4-dynamic.zip

Unzip this and copy Mapbox.framework in following folder:

**$app_path**/node_modules/react-native-mapbox-gl/ios

So that the folder looks like this:
[...]

PR #368, opened 2016-06-21
fix: when mapbox.framework not found (#350)

yutin1987:
Mapbox/Mapbox.h file not found when Mapbox.framework download failed


tmcw:
👍 Thanks!

PR #369, opened 2016-06-23
fix(event): fix #356 onRegionChange is not getting called

### LINKED ISSUE #356: [Android] onRegionChange is not getting called when map is panned or zoomed
The functionality works fine on iOS, but on android, `onRegionChange` never seems to fire off when you drag or zoom the map. I've done some digging and have confirmed that `onMapChanged` does get called within `ReactNativeMapboxGLManager.java` when you drag & zoom (https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java#L233), but it doest seem to be triggering the event within React.

yutin1987:
http://facebook.github.io/react-native/docs/native-components-android.html#events

but for "custom events" (events that are not pre-defined likes onLoad, onScroll, etc..) you will also need to override getExportedCustomDirectEventTypeConstants.

as this line.
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/viewpager/ReactViewPagerManager.java#L58

### DIFF
--- a/android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
+++ b/android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
@@ -1,6 +1,8 @@
 
 package com.mapbox.reactnativemapboxgl;
 
+import java.util.Map;
+
 import android.graphics.Color;
 import android.util.Log;
 import android.os.StrictMode;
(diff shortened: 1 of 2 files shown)


dapetcu21:
I also fixed this in #365. I actually had to throttle these events as Mapbox would fire them at each frame while I was panning the map and the JS thread would get drowned by them. @tmcw

PR #373, opened 2016-07-12
fix #170: preinstall downloads iOS SDK only on Mac. Adds Windows support

### LINKED ISSUE #170: Change the preinstall and start script to something less *nix based
With React Native going to android the use of bash scripts becomes problematic for users on Windows. For example on my windows machine I can't get NPM to install mapbox gl because it attempts to execute the bash script in the /scripts directory which obviously will fail.

tsemerad:
This change allows the package to be installed on Windows, since Windows would choke up on using the `.sh` download script. Also, since it now only downloads the iOS SDK when on Mac, it doesn't unnecessarily install the SDK when on Linux.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "url": "https://github.com/mapbox/react-native-mapbox-gl"
   },
   "scripts": {
-    "preinstall": "./scripts/download-mapbox-gl-native-ios.sh 3.2.0",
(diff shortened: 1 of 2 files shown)


tsemerad:
This fixes #170

tmcw:
👍 Thanks!

Issue #377, opened 2016-07-21
[iOS] Map is not created if no annotations exist on the initial subview layout of RCTMapboxGL


abeck99:
updateMap is called from layoutsubviews, which will in turn create the map if it doesn't exist. This is because the map view needs the bounds from RCTMapboxGL before it's created.

However updateMap is only called if there are no annotations (not sure why). If annotations exist before the initial subview layout, then the map needs to be created anyway.

dapetcu21:
This got fixed in #365. That whole piece of code got refactored.

PR #378, opened 2016-07-21
Create map if no annotations exist on first subview layout

abeck99:
Reference this issue: https://github.com/mapbox/react-native-mapbox-gl/issues/377


dapetcu21:
`-updateMap` doesn't exist anymore. That whole code structure is gone after #365 and this use case should work now. @tmcw

PR #407, opened 2016-08-23
Expose pitchEnabled property

benvium:
By default the user can drag with two fingers (iOS) or three fingers (Android) to adjust the tilt of the map. This PR adds the ability to turn that off.

Added to iOS and Android, and also added to the docs.

NOTE: also fixes a small issue where on iOS `scrollEnabled` and `zoomEnabled` couldn't be changed at run-time. Developer forgot to assign the new values in the setters.

Example:

``` js
 <MapView
      ...
      pitchEnabled={false}
      ...
      />
```


bsudekum:
Looks good, thanks!

PR #409, opened 2016-08-31
Removing iOS in the title

jayfunk:
API Docs should now be merged between the two platforms. Removing the iOS designation from the title. Could make new users confused.


bsudekum:
Good catch, thanks!

PR #412, opened 2016-09-06
fix large polyline annotations on android

### LINKED ISSUE #394: Crash when adding large polyline annotations
I am creating an app where I need to draw driving directions over road networks. I've been using annotations to draw the routes, but I am experiencing a crash without an error message if my route annotation is too large (typically occurs at over 3000 points). I've solved my particular problem by splitting the points up into smaller chunks and drawing many smaller annotations instead of one large annotation, but performance is still less than ideal (the component hangs while drawing the annotations). I wonder if there's a better way of drawing driving directions (or any large polylines) on the map? [How is it done in the Java SDK itself](https://www.mapbox.com/directions/)? Should some directions interface be implemented in this library?

I've put up a repo where you can reproduce the problem: https://github.com/RagnarHal/MapboxLargeAnnotationExample

carlcraig:
This fixes #394 where large polyline annotations on android would cause a crash.

### DIFF
--- a/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
+++ b/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
@@ -9,6 +9,7 @@ import android.graphics.drawable.Drawable;
 import android.support.v4.content.ContextCompat;
 
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.ReadableArray;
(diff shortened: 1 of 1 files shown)


carlcraig:
I have updated this fixing a missing import statement

dapetcu21:
This looks good. Have you tested it?

carlcraig:
Yep, have tested it via npm installing my fork/branch

```
npm i carlcraig/react-native-mapbox-gl#fix-android-large-annotations
```

All seems to work well now.

PR #413, opened 2016-09-07
Feature: Get bounds for positions

mrooding:
# Get bounds for positions

I've added a method which gets the bounding rectangle in GPS coordinates for an array of positions.
## Why?

I have a use case where I have a set of markers on the map and I want to adjust the camera position and zoom level to display all the markers + the user's current location on the map. 

At first, I created a high-level method which determined the bounding box using `LatLngBounds` but also used the `view.setCameraUpdate` method. I eventually decided to implement a more low-level method which focuses just on calculating and returning the bounds. This can be combined with the already available method `setVisibleCoordinateBounds`. 
## Feedback

I'd like to get feedback on the implemented method first before I add it to the documentation. I'd also require some help with the iOS implementation since I'm not familiar with Objective-C.

### DIFF
--- a/android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
+++ b/android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLManager.java
@@ -182,6 +182,7 @@ public class ReactNativeMapboxGLManager extends SimpleViewManager<ReactNativeMap
     public static final int COMMAND_SET_VISIBLE_COORDINATE_BOUNDS = 6;
     public static final int COMMAND_SELECT_ANNOTATION = 7;
     public static final int COMMAND_SPLICE_ANNOTATIONS = 8;
+    public static final int COMMAND_GET_BOUNDS_FOR_POSITIONS = 9;
 
     @Override
     public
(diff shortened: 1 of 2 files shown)


dapetcu21:
Why not implement this in javascript? Naive implementation off the top of my mind:

``` js
const getBoundsForPositions = positions =>
  positions.reduce(
    ([latS, lngW, latN, lngE], [lat, lng]) => 
      ([min(latS, lat), min(lngW, lng), max(latE, lat), max(lngN, lng)]),
    [-Infinity, -Infinity, Infinity, Infinity]
  )
```

mrooding:
It does add a bit more logic to the client but it probably makes sense because I also couldn't find an equivalent for the `LatLngBounds.Builder` in the iOS SDK. I eventually ended up with the following method:

```
const boundaries = positions.reduce((bounds, position) => {
      return [
        bounds[0] ? Math.min(bounds[0], position[0]) : position[0],
        bounds[1] ? Math.min(bounds[1], position[1]) : position[1],
        bounds[2] ? Math.max(bounds[2], position[0]) : position[0],
        bounds[3] ? Math.max(bounds[3], position[1]) : position[1],
      ]
    })
```

Using `[-Infinity, -Infinity, Infinity, Infinity]` will not work if you're only using positive latitude and longitude values since `Infinity` is set to 0 in Javascript and thus 0 is always smaller than any of the positions you're trying to reduce.

PR #414, opened 2016-09-09
Use class methods instead of static properties

mrooding:
The `this` reference used in the static properties like `_onNativeComponentMount` and `_onRegionDidChange` is undefined when using it in combination with the react-relay and babel-relay-plugin (both version 0.9.2). Before adding react-relay to our project everything was working fine but as soon as we switched to Relay with the use of the babel-relay-plugin the errors started occurring as soon as the MapView is loaded.

The `babel-preset-react-native` plugin that we use (version 1.9.0) does support the babel transformer `transform-class-properties` which should take care of `this` bindings in static properties but it does not work for files residing in the node_modules folder. 

The solution was to change all the static properties to methods and add a constructor to do the `this` binding manually. Another solution would be to distribute a "babelified" version of the index.js when publishing to npm as is described [here](http://blog.xebia.com/publishing-es6-code-to-npm/). I'm willing to create a PR to support this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -160,6 +160,22 @@ function addOfflineErrorListener(handler) {
 }
 
 class MapView extends Component {
+  constructor(props) {
+    super(props);
+
+    this._onRegionDidChange = this._onRegionDidChange.bind(this);
+    this._onRegionWillChange = this._onRegionWillChange.bind(this);
+    this._onOpenAnnotation = this._onOpenAnnotation.bind(this);
+    this._onRightAnnotationTapped = this._onRightAnnotationTapped.bind(this);
+    this._onChangeUserTrackingMode = this._onChangeUserTrackingMode.bind(this);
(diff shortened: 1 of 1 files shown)


dapetcu21:
Hmm. I'm not sure about babelifying it before publishing. What happens, for example, if react-native's packager compiles `react` without `babel-plugin-transform-es2015-classes` and we use it in RNMGL? We end up with possibly incompatible classes implementations.

We have to match the version of `babel-preset-react-native` that RN's packager uses at all times if we decide to publish our code compiled. I say better to merge this PR and let RN's packager run babel.

mrooding:
I agree that the little bit of added boilerplate introduced is indeed better than babelifying. Thanks for getting back to me and merging the PR!

dapetcu21:
Sorry for answering late. Crazy months.

Issue #417, opened 2016-09-20
Installing via Cocoapods doesn't link the framework properly


joemckie:
When installing via Cocoapods, I still had to follow the manual installation and link the headers / framework to the project.

Was I installing it wrong, or is this a bug? I'll try on a fresh installation ASAP.

joemckie:
I got some time to take a look into whether this applies to a fresh installation, and it does. I get the following error on build:

```
Undefined symbols for architecture x86_64:
  "_convertToMGLAnnotation", referenced from:
      ___65-[RCTMapboxGLManager spliceAnnotations:deleteAll:toDelete:toAdd:]_block_invoke in libRCTMapboxGL.a(RCTMapboxGLManager.o)
  "_imageFromSource", referenced from:
      -[RCTMapboxGL mapView:imageForAnnotation:] in libRCTMapboxGL.a(RCTMapboxGL.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Setup

• Node: 6.5.0
• Ruby ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]

1ec5:
Duplicate of #350.

joemckie:
So is there a solution? Or is installing via CocoaPods broken?

vjpr:
@joemckie: I have the exact same error (except for architecture arm64) - how did you fix?

brianjd:
@1ec5 How is this a duplicate of #350?

vjpr:
@brianjd I think this is the fix: https://github.com/mapbox/react-native-mapbox-gl/issues/350#issuecomment-257099878

joemckie:
@vjpr I didn't really fix it with CocoaPods, I just followed the manual installation instructions instead

brianjd:
I ended up going the manual route as well.

PR #419, opened 2016-09-21
Adding queryRenderedFeatures on iOS

tsemerad:
I'm trying to make a PR adding `queryRenderedFeatures` support (or `visibleFeaturesAtPoint` and `visibleFeaturesInRect` on iOS), and I need some feedback.

I have the bridge set up to properly send the screen coordinate from JS to Obj C. However, it's failing on line 547 of RCTMapboxGL.m with `Exception thrown while executing UI block: -[MGLMapView visibleFeaturesAtPoint:]: unrecognized selector sent to instance`. I can't figure out why it's giving me this issue. I apologize if this is a simple issue I've overlooked - this is my first time writing React Native bridge code.

I also want feedback on the name and signature on the method. On iOS, it's called `visibleFeaturesAtPoint` and `visibleFeaturesInRect`, and on Android it's called `queryRenderedFeatures`. I'm planning on calling it `queryRenderedFeatures`, and allow passing in either a point or a rect.

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGLManager.m
+++ b/ios/RCTMapboxGL/RCTMapboxGLManager.m
@@ -605,4 +605,31 @@ RCT_EXPORT_METHOD(selectAnnotation:(nonnull NSNumber *) reactTag
     }];
 }
 
+RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
+                  options:(NSDictionary *)options
+                  callback:(RCTResponseSenderBlock)callback)
+{
+    [_bridge.uiManager addUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber *, RCTMapboxGL *> *viewRegistry) {
+        RCTMapboxGL *mapView = viewRegistry[reactTag];
+        if ([mapView isKindOfClass:[RCTMapboxGL class]]) {
+            NSNumber *screenCoordX = options[@"screenCoordX"];
+            NSNumber *screenCoordY = options[@"screenCoordY"];
+            if (!screenCoordX || !screenCoordY) {
+                // TODO: figure out best way to error this. Should we be using RCTPromise rather than a callback?
+//                reject(@"invalid_arguments", @"queryRenderedFeatures(): screenCoordX and screenCoordY are required.", nil);
+                return;
+            }
+            CGPoint point = CGPointMake(screenCoordX.floatValue, screenCoordY.floatValue);
+
+            NSArray * features = [mapView visibleFeaturesAtPoint:point];
+
+            // TODO: also accept Rect to use [mapView visibleFeaturesInRect:]
+
+            // TODO: convert features to JS friendly objects
+
+            callback(features);
+        }
+    }];
+}
+
 @end
--- a/ios/RCTMapboxGL/RCTMapboxGLManager.m
+++ b/ios/RCTMapboxGL/RCTMapboxGLManager.m
@@ -614,34 +614,42 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
         RCTMapboxGL *mapView = viewRegistry[reactTag];
         if ([mapView isKindOfClass:[RCTMapboxGL class]]) {
             NSDictionary *pointDict = options[@"point"];
-            // TODO: also accept Rect to use [mapView visibleFeaturesInRect:]
-            NSDictionary *rect = options[@"rect"];
-            if (!pointDict && !rect) {
+            NSDictionary *rectDict = options[@"rect"];
+            if ((!pointDict && !rectDict) || (pointDict && rectDict)) {
                 reject(@"invalid_arguments", @"queryRenderedFeatures(): one of 'point' or 'rect' is required.", nil);
                 return;
             }
-            NSNumber *screenCoordX = pointDict[@"screenCoordX"];
-            NSNumber *screenCoordY = pointDict[@"screenCoordY"];
+
+            NSArray<id<MGLFeature>> *features;
             NSArray<NSString *> *styleLayerIdentifiersArray = options[@"layers"];
             NSSet<NSString *> *styleLayerIdentifiers;
             if (styleLayerIdentifiersArray) {
                 styleLayerIdentifiers = [NSSet setWithArray:styleLayerIdentifiersArray];
             }
 
-            CGPoint point = CGPointMake(screenCoordX.floatValue, screenCoordY.floatValue);
-
-            NSArray *features = [mapView visibleFeaturesAtPoint:point inStyleLayersWithIdentifiers:styleLayerIdentifiers];
+            if (pointDict) {
+                NSNumber *screenCoordX = pointDict[@"screenCoordX"];
+                NSNumber *screenCoordY = pointDict[@"screenCoordY"];
+                CGPoint point = CGPointMake(screenCoordX.floatValue, screenCoordY.floatValue);
+                features = [mapView visibleFeaturesAtPoint:point inStyleLayersWithIdentifiers:styleLayerIdentifiers];
+            } else {
+                NSNumber *left = rectDict[@"left"];
+                NSNumber *top = rectDict[@"top"];
+                NSNumber *right = rectDict[@"right"];
+                NSNumber *bottom = rectDict[@"bottom"];
+                CGFloat width = right.floatValue - left.floatValue;
+                CGFloat height = bottom.floatValue - top.floatValue;
+                CGRect rect = CGRectMake(left.floatValue, top.floatValue, width, height);
+                features = [mapView visibleFeaturesInRect:rect inStyleLayersWithIdentifiers:styleLayerIdentifiers];
+            }
 
             NSMutableArray *geoJSONFeatures = [[NSMutableArray alloc] init];
             for (id <MGLFeature> feature in features) {
                 NSDictionary *geoJSONGeometry = [self geoJSONGeometryFromMGLFeature:feature];
-
-                NSDictionary *geoJSON = @{
-                        @"type": @"Feature",
-                        @"id": feature.identifier ? feature.identifier : [NSNull null],
-                        @"properties": feature.attributes,
-                        @"geometry": geoJSONGeometry
-                };
+                NSDictionary *geoJSON = @{ @"type": @"Feature",
+                                           @"id": feature.identifier ? feature.identifier : [NSNull null],
+                                           @"properties": feature.attributes,
+                                           @"geometry": geoJSONGeometry };
                 [geoJSONFeatures addObject:geoJSON];
             }
 
@@ -661,10 +669,8 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
         for (MGLShape <MGLFeature> *shape in shapeCollection.shapes) {
             [geometries addObject:[self geoJSONGeometryFromMGLFeature:shape]];
         }
-        return @{
-                @"type": geometryType,
-                @"geometries": geometries
-        };
+        return @{ @"type": geometryType,
+                  @"geometries": geometries };
     }
 
     NSMutableArray *coordinates = [[NSMutableArray alloc] init];
@@ -683,7 +689,7 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
     } else if ([feature isKindOfClass:[MGLMultiPolylineFeature class]]) {
         geometryType = @"MultiLineString";
         MGLMultiPolylineFeature *multiPolyline = (MGLMultiPolylineFeature *)feature;
-        for (MGLPolyline *polyline in multiPolyline.polylines) {
+        for (MGLPolylineFeature *polyline in multiPolyline.polylines) {
             [coordinates addObject:[self getMGLPolylineCoordinates:polyline]];
         }
     } else if ([feature isKindOfClass:[MGLMultiPolygonFeature class]]) {
@@ -702,10 +708,8 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
         }
     }
 
-    return @{
-            @"type": geometryType,
-            @"coordinates": coordinates
-    };
+    return @{ @"type": geometryType,
+              @"coordinates": coordinates };
 }
 
 - (NSMutableArray *)getMGLPolylineCoordinates:(MGLPolylineFeature *)polyline
--- a/ios/RCTMapboxGL/RCTMapboxGLManager.m
+++ b/ios/RCTMapboxGL/RCTMapboxGLManager.m
@@ -607,16 +607,17 @@ RCT_EXPORT_METHOD(selectAnnotation:(nonnull NSNumber *) reactTag
 
 RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
                   options:(NSDictionary *)options
-                  callback:(RCTResponseSenderBlock)callback)
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
 {
     [_bridge.uiManager addUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber *, RCTMapboxGL *> *viewRegistry) {
         RCTMapboxGL *mapView = viewRegistry[reactTag];
         if ([mapView isKindOfClass:[RCTMapboxGL class]]) {
             NSDictionary *pointDict = options[@"point"];
+            // TODO: also accept Rect to use [mapView visibleFeaturesInRect:]
             NSDictionary *rect = options[@"rect"];
-            if (!pointDict) {
-                // TODO: figure out best way to error this. Should we be using RCTPromise rather than a callback?
-//                reject(@"invalid_arguments", @"queryRenderedFeatures(): screenCoordX and screenCoordY are required.", nil);
+            if (!pointDict && !rect) {
+                reject(@"invalid_arguments", @"queryRenderedFeatures(): one of 'point' or 'rect' is required.", nil);
                 return;
             }
             NSNumber *screenCoordX = pointDict[@"screenCoordX"];
@@ -631,14 +632,6 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
 
             NSArray *features = [mapView visibleFeaturesAtPoint:point inStyleLayersWithIdentifiers:styleLayerIdentifiers];
 
-            NSDictionary *geoJSONTypesByMGLClassName = @{
-                    @"MGLPointFeature": @"Point",
-                    @"MGLPolylineFeature": @"LineString",
-                    @"MGLPolygonFeature": @"Polygon"
-            };
-
-            // TODO: also accept Rect to use [mapView visibleFeaturesInRect:]
-
             NSMutableArray *geoJSONFeatures = [[NSMutableArray alloc] init];
             for (id <MGLFeature> feature in features) {
                 NSDictionary *geoJSONGeometry = [self geoJSONGeometryFromMGLFeature:feature];
@@ -652,7 +645,7 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
                 [geoJSONFeatures addObject:geoJSON];
             }
 
-            callback(@[geoJSONFeatures]);
+            resolve(geoJSONFeatures);
         }
     }];
 }
--- a/ios/RCTMapboxGL/RCTMapboxGLManager.m
+++ b/ios/RCTMapboxGL/RCTMapboxGLManager.m
@@ -650,7 +650,7 @@ RCT_EXPORT_METHOD(queryRenderedFeatures:(nonnull NSNumber *)reactTag
                     coordinates = (NSMutableArray *) @[@(feature.coordinate.longitude), @(feature.coordinate.latitude)];
                 } else if ([feature isKindOfClass:[MGLPolylineFeature class]]) {
                     geometryType = @"LineString";
-                    MGLPolyline *polyline = (MGLPolyline *)feature;
+                    MGLPolylineFeature *polyline = (MGLPolylineFeature *)feature;
                     for (int index = 0; index < polyline.pointCount; index++) {
                         CLLocationCoordinate2D coord = polyline.coordinates[index];
                         [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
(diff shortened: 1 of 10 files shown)


1ec5:
> I have the bridge set up to properly send the screen coordinate from JS to Obj C. However, it's failing on line 547 of RCTMapboxGL.m with `Exception thrown while executing UI block: -[MGLMapView visibleFeaturesAtPoint:]: unrecognized selector sent to instance`.

That’s strange. The method definitely is [implemented](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.3.4/platform/ios/src/MGLMapView.mm#L4306). What version of Xcode are you using? This is the SDK’s first method that [uses](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.3.4/platform/ios/src/MGLMapView.h#L1103) the new `NS_SWIFT_NAME` attribute, so if you’re on Xcode 7.2 or lower, it might not be possible to call this method directly.

> I'm planning on calling it `queryRenderedFeatures`, and allow passing in either a point or a rect.

That’s fine. The iOS naming is specific to Cocoa naming conventions, where a getter method without side effects would to be named without a verb.

tsemerad:
I'm using XCode 7.3.1. I tested accessing other attributes and methods on `_map`, and they all worked.

tsemerad:
I got it working now. I just had to remove and re-add Mapbox.framework from the XCode project, just how I originally had [according to the manual installation instructions](https://github.com/mapbox/react-native-mapbox-gl/blob/master/ios/install.md). Now it's recognizing the method properly. Is this something we'll need to warn people about?

1ec5:
> I just had to remove and re-add Mapbox.framework from the XCode project, just how I originally had [according to the manual installation instructions](https://github.com/mapbox/react-native-mapbox-gl/blob/master/ios/install.md).

Maybe that’s related to #417?

tsemerad:
I have it working on iOS by passing in an `options` dictionary with a `point` parameter screen coordinate and optional `layers` array of layer ids. Converting the queried features to GeoJSON might need a little more work - it's working for Point, LineString, and Polygon, but haven't implemented MultiLineString or MultiPolygon. If there's already a known way to get GeoJSON from MGLFeatures, let me know. I see that [mapbox-gl-native #6302](https://github.com/mapbox/mapbox-gl-native/issues/6302) would provide a convenient method.

The question is, what do we want the `queryRenderedFeatures` signature to look like?

This is what I currently have, which I think needs some improvement:

``` javascript
map.queryRenderedFeatures(options, callback)
// example
map.queryRenderedFeatures({
    point: {
      screenCoordX: 212,
      screenCoordY: 111
    },
    layers: ['building']
  },
  (features) => console.log('do something with GeoJSON features')
)
```

I'm thinking of passing in a coordinate, rather than a screen point, to be more in line with Mapbox GL JS (I'm not sure if that's a concern). I started with a screen point since iOS uses `visibleFeaturesAtPoint`, and Android appears to use a screen point as well, but [Mapbox GL JS doesn't](https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures). It accepts [PointLike](https://www.mapbox.com/mapbox-gl-js/api/#PointLike) input. I'm not sure whether we want to add [point-geometry](https://github.com/mapbox/point-geometry) as a dependency just to make it very similar to Mapbox GL JS.

Inspired by Mapbox GL JS:

``` javascript
map.queryRenderedFeatures(geometry, parameters, callback)
// example with point
map.queryRenderedFeatures([44.57, -123.42], {
    layers: ['building']
  },
  (features) => console.log('do something with GeoJSON features')
)

// example with bounding box
map.queryRenderedFeatures([[44.574, -123.426], [44.576, -123.424]], {
    layers: ['building']
  },
  (features) => console.log('do something with GeoJSON features')
)
```

Another alternative to passing `[[latitudeSW, longitudeSW], [latitudeNE, longitudeNE]]` would be `[latitudeSW, longitudeSW, latitudeNE, longitudeNE]` which matches `bounds` that we get from `map.getBounds`.

One other option, that might be more than we need (but is a bit more explicit):

``` javascript
map.queryRenderedFeatures(options, callback)
// example
// one of 'screenCoord', 'mapCoord', 'rect', and 'boundingBox' are required
map.queryRenderedFeatures({
    screenCoord: {
      x: 212,
      y: 111
    },
    mapCoord: {
      latitude: 44.57,
      longitude: -123.43
    },
    rect: {
      left: 0,
      top: 0,
      right: 10,
      bottom: 10
    },
    boundingBox: {
      southWest: {
        latitude: 44.574,
        longitude: -123.426
      },
      northEast: {
        latitude: 44.576,
        longitude: -123.424
      }
    },
    layers: ['building']
  },
  (features) => console.log('do something with GeoJSON features')
)
```

What do you think? Should we accept screen points, lat long coords, or both?

cc @bsudekum, @dapetcu21, @tmcw

1ec5:
> If there's already a known way to get GeoJSON from MGLFeatures, let me know. I see that [mapbox-gl-native #6302](https://github.com/mapbox/mapbox-gl-native/issues/6302) would provide a convenient method.

mapbox/mapbox-gl-native#6302 would make it possible to create an MGLGeoJSONSource from an array of MGLFeatures, which is essentially opposite the direction you’d like to go. With an MGLGeoJSONSource, you could grab the `geoJSONData`, but round-tripping through MGLGeoJSONSource would be a pretty circuitous route to take for feature querying. mapbox/mapbox-gl-native#5623 (or the SDK-level wrapper around it) would ultimately expose a way to convert an MGLFeature to GeoJSON directly.

In the meantime, you’re on the right track with your conversion implementation. For collection types like MGLMultiPolyline and especially MGLShapeCollection, you’ll want to do the conversion recursively.

> I'm thinking of passing in a coordinate, rather than a screen point, to be more in line with Mapbox GL JS (I'm not sure if that's a concern). I started with a screen point since iOS uses `visibleFeaturesAtPoint`, and Android appears to use a screen point as well, but [Mapbox GL JS doesn't](https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures). It accepts [PointLike](https://www.mapbox.com/mapbox-gl-js/api/#PointLike) input.

Based on GL JS’s [documentation](https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures) and [examples](https://www.mapbox.com/mapbox-gl-js/example/queryrenderedfeatures/), it’s clear `queryRenderedFeatures()` actually takes a screen coordinate or screen bounding box as well. `PointLike` is used for coordinates that aren’t geographic coordinates. You could accept geographic coordinates and convert them to screen coordinates before querying them, but that’ll make less sense once mapbox/mapbox-gl-js#3122 is implemented.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
                 // TODO: figure out best way to error this. Should we be using RCTPromise rather than a callback?
 //                reject(@"invalid_arguments", @"queryRenderedFeatures(): screenCoordX and screenCoordY are required.", nil);
                 return;
             }
+            NSNumber *screenCoordX = pointDict[@"screenCoordX"];
+            NSNumber *screenCoordY = pointDict[@"screenCoordY"];
+            NSSet<NSString *> *styleLayerIdentifiers = options[@"layers"];
+
+            CLLocationCoordinate2D location = [_map convertPoint:[sender locationInView:_map] toCoordinateFromView:_map];
+            CGPoint screenCoord = [sender locationInView:_map];
```
1ec5: These two variables don’t appear to be used anywhere at the moment.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+                    MGLMultiPoint *multiPoint = (MGLMultiPoint *)feature;
+                    for (int index = 0; index < multiPoint.pointCount; index++) {
+                        CLLocationCoordinate2D coord = multiPoint.coordinates[index];
+                        [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+                    }
+                    geometryType = @"Polygon";
+                }
+                // TODO: checks for MGLMultiPolyline and MGLMultiPolygon
+
+                NSDictionary *geoJSON = @{
```
1ec5: Also set `id` to the value of the MGLFeature’s `identifier` property. (We don’t say “ID” much in Objective-C since `id` is a keyword.)

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+                    geometryType = @"LineString";
+                } else if ([feature isKindOfClass:[MGLPolygon class]]) {
+                    MGLMultiPoint *multiPoint = (MGLMultiPoint *)feature;
+                    for (int index = 0; index < multiPoint.pointCount; index++) {
+                        CLLocationCoordinate2D coord = multiPoint.coordinates[index];
+                        [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+                    }
+                    geometryType = @"Polygon";
+                }
+                // TODO: checks for MGLMultiPolyline and MGLMultiPolygon
```
1ec5: Also MGLMultiPoint and MGLShapeCollection. Note that MGLPolyline and MGLPolygon currently inherit from MGLMultiPoint (mapbox/mapbox-gl-native#5201), so only handle MGLMultiPoint after those two classes.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+            // TODO: switch to using MGLFeature's geoJSONFeature method if implemented - https://github.com/mapbox/mapbox-gl-native/issues/6302
+            NSMutableArray *geoJSONFeatures = [[NSMutableArray alloc] init];
+            for (id <MGLFeature> feature in features) {
+//                NSString *geometryType = geoJSONTypesByMGLClassName[NSStringFromClass([feature class])];
+                NSString *geometryType;
+                NSMutableArray *coordinates = [[NSMutableArray alloc] init];
+                if ([feature isKindOfClass:[MGLPointFeature class]]) {
+                    geometryType = @"Point";
+                    coordinates = (NSMutableArray *) @[@(feature.coordinate.longitude), @(feature.coordinate.latitude)];
+                } else if ([feature isKindOfClass:[MGLPolyline class]]) {
```
1ec5: Since you’re checking for MGLPointFeature rather than MGLPointAnnotation in the first case above, it’d be more consistent to check for MGLPolylineFeature, MGLPolygonFeature, etc.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+                        [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+                    }
+                    geometryType = @"LineString";
+                } else if ([feature isKindOfClass:[MGLPolygon class]]) {
+                    MGLMultiPoint *multiPoint = (MGLMultiPoint *)feature;
+                    for (int index = 0; index < multiPoint.pointCount; index++) {
+                        CLLocationCoordinate2D coord = multiPoint.coordinates[index];
+                        [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+                    }
+                    geometryType = @"Polygon";
```
1ec5: A [polygon](http://geojson.org/geojson-spec.html#polygon) can have multiple holes, represented as MGLPolygons inside MGLPolygon’s `interiorPolygons` property.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+                    @"MGLPointFeature": @"Point",
+                    @"MGLPolylineFeature": @"LineString",
+                    @"MGLPolygonFeature": @"Polygon"
+            };
 
             // TODO: also accept Rect to use [mapView visibleFeaturesInRect:]
 
-            // TODO: convert features to JS friendly objects
+            // TODO: switch to using MGLFeature's geoJSONFeature method if implemented - https://github.com/mapbox/mapbox-gl-native/issues/6302
+            NSMutableArray *geoJSONFeatures = [[NSMutableArray alloc] init];
```
1ec5: Nit: you can also write this more succinctly as `[NSMutableArray array]`. As a possibly premature but reasonable optimization, you can also specify the size upfront: `[NSMutableArray arrayWithCapacity:features.count]`.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -612,22 +612,76 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+
+
             CGPoint point = CGPointMake(screenCoordX.floatValue, screenCoordY.floatValue);
 
-            NSArray * features = [mapView visibleFeaturesAtPoint:point];
+            NSArray *features;
+            if (!styleLayerIdentifiers) {
+                features = [mapView visibleFeaturesAtPoint:point];
+            } else {
+                features = [mapView visibleFeaturesAtPoint:point inStyleLayersWithIdentifiers:styleLayerIdentifiers];
```
1ec5, line 636: You can call this method for both cases. Pass in `nil` as the second parameter to get the same behavior as `-visibleFeaturesAtPoint:`.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -605,4 +605,85 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+            NSDictionary *pointDict = options[@"point"];
+            NSDictionary *rect = options[@"rect"];
+            if (!pointDict) {
+                // TODO: figure out best way to error this. Should we be using RCTPromise rather than a callback?
+//                reject(@"invalid_arguments", @"queryRenderedFeatures(): screenCoordX and screenCoordY are required.", nil);
+                return;
+            }
+            NSNumber *screenCoordX = pointDict[@"screenCoordX"];
+            NSNumber *screenCoordY = pointDict[@"screenCoordY"];
+            NSSet<NSString *> *styleLayerIdentifiers = options[@"layers"];
```
1ec5: Wouldn’t this be an NSArray rather than an NSSet, given the JavaScript representation? If this works, it’s only because the SDK happens to be calling methods common to both classes. Probably better to convert it into an NSArray using `allObjects`.

1ec5 (changes requested):

ios/RCTMapboxGL/RCTMapboxGLManager.m
tsemerad: You're right. I should be pulling it out of `options` as an `NSArray`, then convert it to an `NSSet`. I just pushed some commits with that change.

> Probably better to convert it into an NSArray using `allObjects`.

You mean convert the NSArray into an NSSet as expected by `visibleFeaturesAtPoint`? Let me know if my most recent change to it looks good. I really appreciate your feedback!

tsemerad:
I think the PR is ready, at least on iOS. I've made all of the requested changes, got `queryRenderedFeatures` working with both `point` and `rect` input, convert the `MGLFeature`s into the 7 various GeoJSON feature types, and updated the API docs. Let me know what you think!

For Android, I've started looking into implementing it. Since `queryRenderedFeatures` is available in the latest Android SDK beta, I'm assuming we want to wait until it lands in a production release first, right? I can make a separate PR then.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -605,4 +605,143 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+            CLLocationCoordinate2D coord = multiPoint.coordinates[index];
+            [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+        }
+    }
+
+    return @{ @"type": geometryType,
+              @"coordinates": coordinates };
+}
+
+- (NSMutableArray *)getMGLPolylineCoordinates:(MGLPolylineFeature *)polyline
```
1ec5: An Objective-C method name should only begin with "get" if the method fetches a resource from the network or something of that nature.

A cleaner approach would be to add a category to MGLPolylineFeature with a method called `-coordinateArray`.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -605,4 +605,143 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+        }
+        return @{ @"type": geometryType,
+                  @"geometries": geometries };
+    }
+
+    NSMutableArray *coordinates = [[NSMutableArray alloc] init];
+
+    if ([feature isKindOfClass:[MGLPointFeature class]]) {
+        geometryType = @"Point";
+        coordinates = (NSMutableArray *) @[@(feature.coordinate.longitude), @(feature.coordinate.latitude)];
```
1ec5: It isn't causing any problems now, but it's unsafe to cast this NSArray to an NSMutableArray. You could declare coordinates as an NSArray without initializing it, then initialize a local NSMutableArray in each if block below where you need one. Not quite as concise, but also reduces the risk of a crash when someone later tries to treat the mutable array as an actual mutable array down below.

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -605,4 +605,143 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+        }
+    } else if ([feature isKindOfClass:[MGLMultiPointFeature class]]) {
+        // this is checked last since MGLPolyline and MGLPolygon inherit from MGLMultiPoint
+        geometryType = @"MultiPoint";
+        MGLMultiPointFeature *multiPoint = (MGLMultiPointFeature *)feature;
+        for (int index = 0; index < multiPoint.pointCount; index++) {
+            CLLocationCoordinate2D coord = multiPoint.coordinates[index];
+            [coordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+        }
+    }
```
1ec5, line 711: What about MGLShapeCollectionFeature (FeatureCollection)?

ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -605,4 +605,143 @@ - (NSArray*)serializePacksArray:(NSArray<MGLOfflinePack*>*)packs
...
+
+    NSMutableArray *outerRingCoordinates = [[NSMutableArray alloc] init];
+    for (int index = 0; index < polygon.pointCount; index++) {
+        CLLocationCoordinate2D coord = polygon.coordinates[index];
+        [outerRingCoordinates addObject:@[@(coord.longitude), @(coord.latitude)]];
+    }
+    [coordinates addObject:outerRingCoordinates];
+
+    for (MGLPolygonFeature *interiorRing in polygon.interiorPolygons) {
+        NSMutableArray *interiorRingCoordinates = [[NSMutableArray alloc] init];
```
1ec5: This is fine, although you could also call the current method recursively here.

1ec5 (changes requested):
This is looking pretty good. A couple code style issues and a missing case.

ios/RCTMapboxGL/RCTMapboxGLManager.m
1ec5: Yikes, never mind, I missed that it's right up there at the top.

ios/RCTMapboxGL/RCTMapboxGLManager.m
tsemerad: How about not creating an NSArray in the first place? I'm thinking replacing line 680 with the following:

``` objective-c
coordinates = [[NSMutableArray alloc] initWithArray:@[@(feature.coordinate.longitude), @(feature.coordinate.latitude)]];
```

1ec5: That works too. (`@[]` creates an NSArray, after all.)

ios/RCTMapboxGL/RCTMapboxGLManager.m
tsemerad: Thanks. I've created a category for MGLPolylineFeature and MGLPolygonFeature, each in their own files. Do you have a preferred naming convention? Should I call it `MGLPolylineFeature+coordinateArray.m`, or maybe `MGLPolylineFeature+RCTMapboxGL.m`?

1ec5: How about: MGL{Polyline,Polygon}Feature+RCTAdditions.{h,m}

ios/RCTMapboxGL/RCTMapboxGLManager.m
tsemerad: Wouldn't calling it recursively nest the arrays too much?

```
[outerRing, interiorRing1, interiorRing2, ...]
```

vs

```
[outerRing, [interiorRing1], [interiorRing2], ...]
```

I've left it as is in my refactoring to using a category.

tsemerad:
I pushed a new commit addressing your concerns. I put the category on MGLPolyline and MGLPolygon, rather than their "Feature" counterparts, since MGLMultiPolyline's `polylines` attribute (and MGLMultiPolygon's `polygons` attribute) return them. I had overlooked that originally. Thanks again for the great feedback.

1ec5 (approved):
This looks 👌. There are some additions opportunities for refactoring, for example moving all the GeoJSON conversion logic into more MGLFeature categories, but that can always happen down the line.

Thanks again for taking up this task. @bsudekum, good to merge?

lyzidiamond:
HI TANNER!!!!!

bsudekum (approved):

1ec5:
Only thing left to do is to resolve the conflicts.

bsudekum:
@tsemerad want to fix the conflicts? then we can merge

bsudekum:
heh, didn't reload the page, 🎉

tsemerad:
Glad I could help. Haha, hi @lyzidiamond! Good to hear from you! You probably didn't expect to see my name pop up in your Github feed.

1ec5:
> mapbox/mapbox-gl-native#5623 (or the SDK-level wrapper around it) would ultimately expose a way to convert an MGLFeature to GeoJSON directly.

We ultimately implemented this in the iOS SDK as [`-[MGLShape geoJSONDataUsingEncoding:]`](https://www.mapbox.com/ios-sdk/api/3.5.0/Classes/MGLShape.html#/c:objc%28cs%29MGLShape%28im%29geoJSONDataUsingEncoding:). #8493 will add a similar API to the Android SDK.

PR #421, opened 2016-09-22
support for custom annotation views

aksonov:
Example of usage:

```
import Mapbox, { MapView, Annotation } from 'react-native-mapbox-gl';


...
render(){
     <MapView><Annotation coordinate={{longitude:LONG, latitude:LAT}}>... custom view here ..</Annotation></MapView>
}
```


ios/RCTMapboxGL/RCTMapboxAnnotation.h
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
```
1ec5: This copyright block is unneeded, right? Or is this adapted from somewhere?

ios/RCTMapboxGL/RCTMapboxAnnotationManager.m
```diff
@@ -0,0 +1,41 @@
...
+
+@end
+
+@implementation RCTMapboxAnnotationManager
+
+RCT_EXPORT_MODULE()
+
+- (UIView *)view
+{
+    RCTMapboxAnnotation *marker = [RCTMapboxAnnotation new];
```
1ec5: Nit: `[[RCTMapboxAnnotation alloc] init]`. `new` works, but it’s an antipattern.

1ec5 (commented):
The Objective-C side looks good to me, other than a couple small things.

1ec5:
@bsudekum or @dapetcu21, could you review the JavaScript side of this change?

MapboxAnnotation.js
```diff
@@ -0,0 +1,283 @@
...
+      default:
+        break;
+    }
+  }
+  
+  render() {
+    let image;
+    if (this.props.image) {
+      image = resolveAssetSource(this.props.image) || {};
+      image = image.uri;
```
tmcw: It looks like a specified but invalid image will fail silently here. Does `resolveAssetSource` emit a warning, or should we emit a warning or exception here?

index.js
```diff
@@ -13,6 +13,7 @@ import {
 import cloneDeep from 'lodash/cloneDeep';
 import clone from 'lodash/clone';
 import isEqual from 'lodash/isEqual';
+import Annotation from './MapboxAnnotation';
```
tmcw: Can we name the class `MapboxAnnotation` or the file `Annotation` to remove this mismatch?

MapboxAnnotation.js
```diff
@@ -0,0 +1,283 @@
...
+  
+  /**
+   * Callback that is called when a drag on this marker finishes. This is usually the point you
+   * will want to setState on the marker's coordinate again
+   */
+  onDragEnd: PropTypes.func,
+};
+
+const defaultProps = {
+  onPress() {},
```
tmcw: looks like an unnecessary trailing `,` here

tmcw (commented):
Overall, looks extremely good.

rmevans9:
Is there a plan to also implement this in Android? I would love to see this functionality work on both platforms.

aksonov:
Thanks, all, fixed.

@rmevans9 Sorry, only iOS at this moment, but you could adopt it easily for Android by looking into react-native-maps Android part (i've studied iOS part, AIRMarker\* files)

aksonov:
I believe it could be merged and then improved gradually. Old annotations (not custom) should work also as before. Btw, i have very strange issue with custom annotations - it works stable on simulator, but sometimes custom annotations are not shown on device. Maybe it is bug with iOS SDK.

1ec5:
> Btw, i have very strange issue with custom annotations - it works stable on simulator, but sometimes custom annotations are not shown on device. Maybe it is bug with iOS SDK.

If you can reproduce this in a straightforward, stripped down project that uses the iOS SDK in Objective-C or Swift but not React Native, please file a bug with the test case in the mapbox/mapbox-gl-native repository. Otherwise, React Native and RNMBGL do a lot of magic that makes it difficult to attribute a bug to the SDK with certainty.

aksonov:
When it could be merged?

P.S. I've fixed issues connected with device - map could be not created during add/remove annotations so, i've added proper fix.

MapboxAnnotation.js
aksonov: fixed

MapboxAnnotation.js
aksonov: fixed

ios/RCTMapboxGL/RCTMapboxAnnotationManager.m
aksonov: fixed

ios/RCTMapboxGL/RCTMapboxAnnotation.h
aksonov: fixed

1ec5:
Are any changes needed to the API documentation? We should probably point out that this feature is only available for iOS at the moment.

aksonov:
@1ec5 ok, added docs

1ec5:
Thanks! 🎉

Android support for view-backed annotations is being added in #423, although that PR takes a markedly different approach. Instead of arbitrary React Native views, the developer specifies the usual annotation options (plus a few more) and those options are implemented using MarkerViews.

aksonov:
@1ec5 I'm preparing some improvements for this PR, because now annotation could disappear after scrolling map away and return back. So each Annotation must have 'id' and be used as reuseIdentifier for views.

aksonov:
I will submit other PR for that.

tlvenn:
@1ec5 Is there any platform reason/restriction as to why the Android support must be different ?

I dont want to dismiss the work @mrooding has put into his PR to bring this feature to android but it seems to me that we should aim to have an api that works the same on both IOS and Android if possible.

mrooding:
@tlvenn I can't agree more. As said in my PR, if anyone could give me some pointers on how to get started I'm more than happy to contribute to the Android solution.

tlvenn:
Unfortunately I cant help you much on this particular issue :(

One thing that I would like to confirm however is that @aksonov implementation allows for dynamic custom annotation views ? Similarly to the `annotations` array, could we have a `viewAnnotations` array we could manipulate ?

One could have some dynamic children but it suddenly force your app to have a different flow than how you would do thing with the `annotations` array.

1ec5:
Unfortunately, I don't know enough about how either the Android SDK or React Native works to give a cogent answer to these questions. All I know is that the Android and iOS SDKs in principle support the same feature: representing point annotations (markers) as native views (marker views); like any ordinary annotation, they can be added, removed, and relocated dynamically.

tlvenn:
I kinda also think that @mrooding approach to leverage existing entry point into the annotations is much better than introducing an entire new way to define some annotations.

Why not simply support the following shape for a view type annotation:

```
annotations: [{
  coordinates: [40.72052634, -73.97686958312988],
  type: 'view',
  data: {}, //any data you need to properly render the view
  renderer: func // renderer function that takes data as param and return a react element
}]
```

 The view would simply delegate its render method to the renderer function passing the data object as param.

Having a single source of truth for annotations is especially important when it comes to adding support for feature such as clustering of annotations and the like. All annotations no matter their type should be in the annotation array imho.

It also force the annotation view components to be stateless which is probably a good idea and aligned to current annotation types which are all stateless by nature.

I also dont think it's necessarily a good idea to introduce support for nested elements with `children` unless we do it consistently such as being able to pass any annotation type as children if this actually deemed a good idea, which i am not convinced.

Being able to back annotation with a view is absolutely fantastic but I hope we take the time to think what is the best approach to expose this to the end user.

@bsudekum @dapetcu21 any though on this ?

1ec5:
> All annotations no matter their type should be in the annotation array imho.

That’s reasonable. However, I think the rationale behind allowing annotation views to be arbitrary children was that there are lots and lots of things you can do with views (at least on iOS). Nothing’s stopping you from embedding a switch control or an arbitrarily nested view hierarchy inside an annotation view, if that’s what you want. Would it be possible to enable all that flexibility inside the existing `annotations` array without adding explicit support for every possible customization? (Not a rhetorical question; I don’t profess to know what the right answer is.)

> pass any annotation type as children

View-backed polylines and polygons aren’t supported by the native SDKs anyways.

wafisher:
First off, I want to say this feature is awesome. I ran into a slight issue using it, though.

React seems to want to layout the `Annotation` with Flexbox in a container the width of the screen. With no extra layout code I have the following:

```
<Annotation id='location' coordinate={{latitude:40.750, longitude:-73.980}}>
  <Icon name="my-location" size={20} />
</Annotation>
```

and React puts this in a container view but on the left side, which is not where that coordinate is on the map. The coordinate is in the center of the _container_ view.

screenshot 1

If I change the code to

```
<Annotation id='location' style={{alignItems: 'center'}} coordinate={{latitude:40.750, longitude:-73.980}}>
  <Icon name="my-location" size={20} />
</Annotation>
```

It at least sits in the center of the container view, where it should be:

screenshot

Sorry the pictures aren't terribly useful since Xcode only shows the wireframe of the map view in the debug view hierarchy.

Anyhow, this fix works but maybe there's a better way. But it should be a good workaround for anyone who needs something now.

aksonov:
Yes, i saw this issue as well and did the same workaround. 
Don’t understand why it happened yet, looks like it is connected with React Views internals. Any help would be appreciated.

PR #422, opened 2016-09-22
[Android] Make remote images for annotations work

### LINKED ISSUE #391: Remote images don't work on Android
On Android, remote images should be loaded on a different thread than the UI, so that the `NetworkOnMainThread` exception doesn't happen.

rmevans9:
I have done some testing and this appears to make remote images on annotations work for my use case at least. I am currently just returning null in the event of an exception - not sure what the preference here would be. I could catch it at the top level and try and pull it from a drawable name possibly but I don't know that makes much sense. Let me know any tweaks you guys want!

Resolves #391

### DIFF
--- a/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
+++ b/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationOptionsFactory.java
@@ -11,7 +11,6 @@ import android.os.AsyncTask;
 import android.util.Log;
 
 import com.facebook.react.bridge.ReadableMap;
-import com.facebook.react.bridge.ReadableArray;
 import com.mapbox.mapboxsdk.annotations.Annotation;
 import com.mapbox.mapboxsdk.annotations.Icon;
 import com.mapbox.mapboxsdk.annotations.IconFactory;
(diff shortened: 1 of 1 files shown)


rmevans9:
Actually hold off on reviewing - it seems my changes did break local images for annotations. Likely because of the additional catches and the fact that loading from drawable was listening for a specific type of exception

rmevans9:
Re-did the exception catching/throwing so local images still work. Note that right now I am catching anything that isn't a IOException and rethrowing it as an IOException since any exception while pulling the image from the web basically ends up in an I/O Exception. If desired I can remove that then I would need to declare all the exceptions that could occur from pulling from the web.

markusekblad:
This is working for me. I'm wondering if it might be leaking memory without a HttpURLConnection disconnect() and InputStream close() tho?

MindFreeze:
Will this be merged soon?

vbuch:
Would love to see this merged. Need it. Badly.

henrikra:
@rmevans9 Have you thought about @markusekblad 's proposal?

rmevans9:
I have not - I moved on from using this library in favor of the alternative so I have not touched this PR for a long time.

henrikra:
Oh okay. Do you use react-native-maps now or?

rmevans9:
Yes with mapbox tiles.

henrikra:
Okay great. So maybe close this PR since here is the newer version of this PR https://github.com/mapbox/react-native-mapbox-gl/pull/581

rmevans9:
👍

PR #426, opened 2016-09-27
add reuseIdentifier usage for custom iOS annotations to fix scroll away disappearing

aksonov:
It appears that reuseIdentifier is mandatory to use custom views (i thought it is optional, just for performance improvements). 
https://github.com/mapbox/mapbox-gl-native/issues/6458#issuecomment-249628078

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxAnnotation.m
+++ b/ios/RCTMapboxGL/RCTMapboxAnnotation.m
@@ -17,4 +17,14 @@
 @implementation RCTMapboxAnnotation {
 }
 
+-(NSString *)reuseIdentifier {
+    return self.id;
+}
+
+-(void)layoutSubviews {
+    [super layoutSubviews];
+    [self.map layoutSubviews];
+}
+
+
 @end
--- a/ios/RCTMapboxGL/RCTMapboxAnnotation.m
+++ b/ios/RCTMapboxGL/RCTMapboxAnnotation.m
@@ -23,7 +23,7 @@
 
 -(void)layoutSubviews {
     [super layoutSubviews];
-    [self.map layoutSubviews];
+    [self.map restoreAnnotationPosition:self.id];
 }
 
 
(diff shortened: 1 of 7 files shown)


1ec5:
@bsudekum for review.

1ec5:
Hi @aksonov, thanks for this PR. It looks like this PR does more than provide the reuse identifier. Can you describe what else has changed?

1ec5:
Ah, I see that these changes were made in response to mapbox/mapbox-gl-native#6474.

ios/RCTMapboxGL/RCTMapboxAnnotation.m
```diff
@@ -17,4 +17,14 @@
 @implementation RCTMapboxAnnotation {
 }
 
+-(NSString *)reuseIdentifier {
+    return self.id;
+}
+
+-(void)layoutSubviews {
+    [super layoutSubviews];
+    [self.map restoreAnnotationPosition:self.id];
```
1ec5, line 26: If the intent is to maintain the view’s center after changing its size, you can instead:
1. Store the current `center` in a local variable.
2. Call `[super layoutSubviews]`.
3. Set the `center` back to the value stored in (1).

aksonov: @1ec5 Unfortunately it is not possible. When layoutSubviews is called, position is already changed by internal react native stuff (when JS changes size of that react view, for example).

mrooding:
It seems that this PR will need to be merged to fix the broken state of master due to #421. It's trying to import from `./Annotation` which was not included in that PR

1ec5:
Since this PR fixes a broken build, I'm going to go ahead and merge it. The manual annotation repositioning is probably unperformant and sort of defeats the purpose of using the SDK's view-backed annotation functionality, but that can be addressed in future PRs.

PR #440, opened 2016-10-17
Exposing min/max zoom level properties to react native

briancalvium:
# Summary

This pull request exposes the min/max zoom limit properties to javascript. The defaults are set to the same as in the native code (unfortunately, I didn't see a clear way to get the defaults from the native code, and had to hard code the default values).
# Test Plan

In an app with a Mapbox MapView component
Add the properties `minimumZoomLevel` and `maximumZoomLevel` with the values 15 and 18.

> Zooming should then be limited to those values.


briancalvium:
Forgot to mention: This is for both iOS and Android.

tmcw:
👍 Thanks @briancalvium !

PR #447, opened 2016-10-26
PropTypes.object() is not a function. Use shape() instead.

wafisher:
This should remove the following error message from the console:

```
<Warning>: Warning: You are manually calling a React.PropTypes validation function for the `undefined` prop on `<<anonymous>>`. This is deprecated and will not work in the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.
```


pauliusuza (approved):

tmcw:
👍 Thanks, merging!

PR #449, opened 2016-10-28
Adding onCloseAnnotation to iOS

wafisher:
Exposing `didDeselectAnnotation` for iOS. A couple open questions:
1. Is there an equivalent Android event? All I could find was a `MarkerClickListener` but no corresponding un-click listener.
2. I assume there's no way to get an event "user selected _another_ annotation" when one is already selected? Right now, the only way to figure that out is to watch for a deselect and select (of another annotation) back to back but I'd obviously prefer not to have to do weird timing in the code like that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,6 +167,7 @@ class MapView extends Component {
     this._onRegionDidChange = this._onRegionDidChange.bind(this);
     this._onRegionWillChange = this._onRegionWillChange.bind(this);
     this._onOpenAnnotation = this._onOpenAnnotation.bind(this);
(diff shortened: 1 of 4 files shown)


tmcw:
Would the Android equivalent be [OnInfoWindowCloseListener?](https://www.mapbox.com/android-sdk/api/4.2.0-beta.4/com/mapbox/mapboxsdk/maps/MapboxMap.OnInfoWindowCloseListener.html)?

tmcw:
Merging this while it's still mergable - Android support for this method can be a todo.

PR #450, opened 2016-11-04
Fix typo in CocoaPods instructions

skylarmb:
Despite the disclaimer that the path depends on your Podfile location, 99% of people will need to change that to be `../node_modules` so why not have that as the default in the instructions


tmcw:
👍 Thanks @skylarmb !

PR #455, opened 2016-11-17
fix podspec

maddanio:


tmcw:
👍  Thanks @maddanio !

maddanio:
actually @vjpr found it (see [issue](https://github.com/mapbox/react-native-mapbox-gl/issues/350)), just seems he forgot to submit a patch :)

PR #467, opened 2016-11-30
Fix bug preventing "Mapbox.getOfflinePacks" from returning on iOS

### LINKED ISSUE #389: getOfflinePacks is not working on ios
hello, after executing "addOfflinePackForRegion" on ios and android, i want to retrieve it somewhere in the app using "getOfflinePacks",

 Mapbox.getOfflinePacks()
      .then(packs => {
        console.log('packs', packs);
      })
      .catch(err => {
        console.error(err); // Handle error
      })

on android, i am getting packs,  but not on ios.

lucaswoj:
@1ec5's [diagnosis and suggested fix](https://github.com/mapbox/react-native-mapbox-gl/issues/389#issuecomment-247823285) is spot on 👍 

fixes #389

cc @1ec5 @bsudekum @dapetcu21

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGLManager.m
+++ b/ios/RCTMapboxGL/RCTMapboxGLManager.m
@@ -153,7 +153,7 @@ RCT_EXPORT_METHOD(setAccessToken:(nonnull NSString *)accessToken)
     _loadedPacks = NO;
     
     // Setup pack array loading notifications
-    [[MGLOfflineStorage sharedOfflineStorage] addObserver:self forKeyPath:@"packs" options:0 context:NULL];
+    [[MGLOfflineStorage sharedOfflineStorage] addObserver:self forKeyPath:@"packs" options:NSKeyValueObservingOptionInitial context:NULL];
     _packRequests = [NSMutableArray new];
     
     // Setup offline pack notification handlers.


ios/RCTMapboxGL/RCTMapboxGLManager.m
```diff
@@ -153,7 +153,7 @@ - (id)init
     _loadedPacks = NO;
     
     // Setup pack array loading notifications
-    [[MGLOfflineStorage sharedOfflineStorage] addObserver:self forKeyPath:@"packs" options:0 context:NULL];
+    [[MGLOfflineStorage sharedOfflineStorage] addObserver:self forKeyPath:@"packs" options:NSKeyValueObservingOptionInitial context:NULL];
```
1ec5, line 156: Does this set up a race condition where the pack notification handlers (set up below) aren't created by the time the initial change notification comes in? It would depend on whether the initial change notification is posted synchronously or whether it is posted asynchronously at the end of the current run loop iteration.

lucaswoj: I have not observed this race condition in practice. Without a deeper understanding of the framework, I can't speak to the theory.

PR #487, opened 2017-01-07
Update headers for RN@0.40.0

sntran:
React Native has made a breaking change to header paths in [0.40.0](https://github.com/facebook/react-native/releases/tag/v0.40.0).

Besides fixing the headers, this adds `RN@0.40` and `React@15.4.0` as
peer dependencies and bump version.

This PR simply adjusts the headers.


bsudekum:
> Besides fixing the headers, this adds RN@0.40 and React@15.4.0 as peer dependencies and bump version.

Is adding react and react native as a peer dependency necessary?

sntran:
It breaks compatibility with older version of React Native. For this particular PR, React Native 0.40 is required, so I added it as peer dependency.

aksonov:
@sntran I'm still having compilation error about <React/RCTConvert+CoreLocation.h> file not found (within RCTMapboxGLConversions.h> with this PR, any other path/settings i should alter?

aksonov:
@sntran Maybe it is because CocoaPods settings were not changed?

sntran:
I have always manually installed for iOS so I don't know about CocoaPods. I'll see what I can do.

aksonov:
@sntran Thanks - right, i've just checked with manual install - it works well! Unfortunately some libraries like latest iOS XMPP Framework have only CocoaPods install option...

aksonov:
Looks like RN0.40 cannot be installed via CocoaPods at all now - 'yoga' library is missed (PR is in progress as i see).

bsudekum:
Thanks! I think the cocoapod failures could be fixed with https://github.com/mapbox/react-native-mapbox-gl/pull/499

ivankennethwang:
Any updates when can we have a new release?

PR #489, opened 2017-01-19
Allow annotation images from the file system

mattvot:
Adds support for using images in the file system as annotation images on both iOS and Android.

### DIFF
--- a/API.md
+++ b/API.md
@@ -267,7 +267,7 @@ mapbox://styles/bobbysud/cigtw1pzy0000aam2346f7ex0
   id, // required. string. Unique identifier used for adding or selecting an annotation.
   annotationImage: { // optional. Marker image for type=point
     source: {
-      uri // required. string. Either remote image URL or the name (without extension) of a bundled image
+      uri // required. string. Either remote image URL, local image file URL or the name (without extension) of a bundled image
     },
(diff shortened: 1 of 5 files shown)


API.md
```diff
@@ -309,7 +309,7 @@ annotations: [{
   title: 'Important',
   subtitle: 'Neat, this is a custom annotation image',
   annotationImage: {
-    source: { uri: 'https://cldup.com/7NLZklp8zS.png' },
+    source: { uri: 'file:///var/mobile/Containers/Data/Application/8821CE3F-C0DA-4B86-B4F7-85DE1E158943/Documents/custom-annotation.png' },
```
benvium, line 314: Perhaps we could say how you get this path in React Native, i.e. via `react-native-fs` or `react-native-fetch-blob`s File access APIs

mattvot: Sure, I can add this description under line 285

henrikra:
@mattvot Wouldn't it be better if I could use local images like everywhere else like `require('path/to/my/image')`?

benvium:
@henrikra That would be good too. This PR is useful for accessing marker images that have been downloaded to the file system (something that's handy for our use of this library here at Calvium).

nitaliano:

PR #490, opened 2017-01-19
Change import to work on case-sensitive MacOS filesystems.

gillessed:
I had a problem setting up RN mapboxgl to work on my system because of
this.


tmcw:
👍 Thanks @gillessed !

PR #499, opened 2017-02-07
fixing podspec for good

maddanio:
this makes the podspec use globbing to make sure it really picks up all source files


maddanio:
...because for me it broke again recently...

PR #505, opened 2017-02-20
Add support for paths with spaces

theill:
If script is being run from a path having spaces it will fail. This patch will ensure to support spaces by encapsulating concatenated script path with double-quotes.


tmcw:
👍 Thanks!

Issue #516, opened 2017-03-13
Update to Android Mapbox GL Native 5.0.0


tobrun:
This ticket covers the actions needed to get react-native-mapbox-gl up-and-working with the latests Android Mapbox SDK version. We recently released the third beta version of 5.0.0 and are planning a stable release in the upcoming weeks.

![device-2017-03-13-081652](https://cloud.githubusercontent.com/assets/2151639/23846380/756129ec-07cd-11e7-8e63-664f6c25d8dd.png)

# Crash from 4.2.0-beta.4 and up

The  https://github.com/mapbox/mapbox-gl-native/issues/7587 upstream ticket in gl-native covers a crash that was added in Android Mapbox SDK 4.2.0-beta and up. Since that version, we started to load the native library as part of the MapboxAccountManager creation. This requires, with current gl-native threading model, that this class must be created from the main thread. The code currently found in the repository doesn't take this in account. Loading classes from the main thread goes for every class that contains the following definition, or classes that call into components using:

```
static {
  System.loadLibrary("mapbox-gl");
}
```

# Adapt API to 5.0.0 changes

## MapboxAccountManager.java to Mapbox.java

With the 5.0.0 release, we are changing `MapboxAccountManager` to `Mapbox` as this component has grown in responsibility in the latest few SDK releases.

```java
// 4.x.x
MapboxAccountManager.start(context, <your access token here>);
// 5.0.0
Mapbox.getInstance(context, <your access token here>));
```

If you need a reference to the set access token later on in the code you can do:

```java
Mapbox.getAccessToken()
```

## AndroidManifest.xml

The telemetry service package has moved. You'll need to change this from:

```java
// 4.x.x
<service android:name="com.mapbox.mapboxsdk.telemetry.TelemetryService"/>
// 5.0.0
<service android:name="com.mapbox.services.android.telemetry.service.TelemetryService"/>
```

## Activity lifecycle events

Part of supporting Android Nougat means we have to also support the new Multi-Window feature. This required us to move some of the logic to the onStart() and onStop() methods. When switching over to 5.0 Android Studio won't complain about these missing but when compiling your application will immediately crash if you don't add them to the activity:

```java
@Override
protected void onStart() {
  super.onStart();
  mapView.onStart();
}

@Override
protected void onStop() {
  super.onStop();
  mapView.onStop();
}
```

## Min/Max zoom level changes

[...]

PR #517, opened 2017-03-13
Fix Android crash from SDK 4.2.0-beta.4 and up

esamelson:
This PR fixes the issue first reported in mapbox/mapbox-gl-native#7587 and identified by @tobrun in #516. We ensure that the creation of the `MapboxAccountManager` class happens on the main thread, which was required starting with Android SDK 4.2.0-beta.4. If merged, this PR will make the current `master` usable on Android once again.


gillessed:
Just tried out these changes. Looks like it works \o/.

bsudekum:
Looks good, thanks!

PR #538, opened 2017-04-20
Allow base64-encoded image data-URI to be used for annotationImage source URI

munkelo:
I'd like to be able to dynamically generate images locally before using them as annotation images, and this seemed like simple way to support this functionality.


nitaliano:

PR #541, opened 2017-04-21
Android custom annotation bugfixes.

gillessed:
- Custom annotations would sometimes not get rendered on the screen.
  Haven't nailed down the exact problem, but it was some sort of
  interleaving of operations from adding children and computing custom
  marker views. Moving the marker view addition to when the map finishes
  loading resolves this.
- The adapter for the react native marker views now handles the caching
  behaviour correctly and re-uses views when they are available.
- Make sure each layout actually gets the height and width of the react
  native views from the correct shadow nodes so that they don't all have
  the same measurements by accident.
- Use custom shadow node to cache view widths and heights correctly so that they don't all have the same dimensions by accident.
- Update locations of annotations when they are changed.
- Split annotation views into a map. The manager is a singleton and all
  the existing map views go through it so the children need to be kep
  separately.
- Safeguard against relayout being called after onDrop(), resulting in a
  null pointer exception.

### DIFF
--- a/API.md
+++ b/API.md
@@ -364,9 +364,6 @@ The `Annotation` view has the following required props:
 2. React Native views do not work with the regular `onAnnotationTapped` on need to
    add their own tap handling (e.g. by using a `TouchableHighlight`).
 
-3. (*Android only*) Adding a view with style `flex: 1` and no `width` or `height` set can cause
-   views to behave a little strangely. It is recommended to add them.
-
 ###### Example
 
 ```
--- a/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationView.java
+++ b/android/src/main/java/com/mapbox/reactnativemapboxgl/RNMGLAnnotationView.java
@@ -1,30 +1,27 @@
 package com.mapbox.reactnativemapboxgl;
 
 import android.content.Context;
-import android.util.Log;
 
-import com.facebook.react.uimanager.LayoutShadowNode;
 import com.facebook.react.views.view.ReactViewGroup;
 import com.mapbox.mapboxsdk.geometry.LatLng;
 
+import java.util.HashSet;
+import java.util.Set;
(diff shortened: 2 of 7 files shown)


gillessed:
@alinapaz Here's all the work we've done on our side for bugfixes. Could you get someone to take a look at this? There are some docs updates which I plan on getting soon as well.

alinapaz:
@gillessed ty for the flag 😄  adding @zugaldia to take a look.

zugaldia:
@gillessed thank you so much for the contribution. I just checked out your branch and everything seems to be running smoothly.

Something I don't see in this PR is any tests. Any chance you could share a minimum amount of code that reproduces the issue that you're trying to solve? It could be as simple as a variation of [`example.js`](https://github.com/mapbox/react-native-mapbox-gl/blob/master/example.js) that doesn't work with `master` but that it does with this branch.

gillessed:
@zugaldia Indeed. We did all of our testing locally. Here's a gist that's similar to `example.js`: https://gist.github.com/gillessed/a1091bee39118021f7a7cb4703cfb7cc. Long press to add custom annotation and play around with them.

zugaldia:
@gillessed thank you for providing the example. I could run it and it all works as expected. I also compared it with the current state of `master` and I can definitely see the issues you mention in the OP. From what I can tell from this example, this PR is 👍 . I'm gonna ahead and merge it.

screenshot_20170426-115246

Finally, quick note for anyone trying to run the test file. I had to make the following minor changes to the `rnmgl-example.jsx` file to make it work (besides renaming it to `index.android.js`):

- Add `AppRegistry` to the `react-native` imports list.
- Replace `@mobile/react-native-mapbox-gl` with `react-native-mapbox-gl` to use the local branch copy within `node_modules`.
- And logically, add a valid token and the actual name of the app in the last line.

PR #555, opened 2017-05-16
fix custom annotation views bug in iOS

### LINKED ISSUE #512: Unrecognized selector with custom annotation
Using mapbox on iOS and have several custom annotations.

When adding an annotation after the map has rendered I get this error:
`[RCTMapboxAnnotation annotationImageSource]: unrecognized selector sent instance ...`

jackfengji:
1. fix #512 crash when custom annotations change
2. fix custom annotations not being removed from map after it's removed
from RN children

### DIFF
--- a/ios/RCTMapboxGL/RCTMapboxGL.m
+++ b/ios/RCTMapboxGL/RCTMapboxGL.m
@@ -45,7 +45,7 @@
     MGLAnnotationVerticalAlignment _userLocationVerticalAlignment;
     /* So we don't fire onChangeUserTracking mode when triggered by props */
     BOOL _isChangingUserTracking;
-    NSMutableDictionary<NSString *, UIView *> *_reactSubviews;
+    NSMutableArray<UIView *> *_reactSubviews;
 }
 
 // View creation
@@ -57,7 +57,7 @@
         _clipsToBounds = YES;
         _finishedLoading = NO;
         _annotations = [NSMutableDictionary dictionary];
-        _reactSubviews = [NSMutableDictionary dictionary];
+        _reactSubviews = [NSMutableArray new];
     }
 
     return self;
@@ -110,10 +110,10 @@
     }
 
     [self addSubview:_map];
-    for (NSString *key in [_reactSubviews allKeys]) {
-        [_map addAnnotation:_reactSubviews[key]];
+    for (UIView *annotation in _reactSubviews) {
+        [_map addAnnotation:(RCTMapboxAnnotation *)annotation];
     }
-    
+
     [self layoutSubviews];
 }
 
@@ -139,8 +139,7 @@
         RCTMapboxAnnotation * annotation = (RCTMapboxAnnotation *) subview;
         annotation.map = self;
         [_map addAnnotation:annotation];
-        NSString *key = annotation.reuseIdentifier;
-        _reactSubviews[key] = annotation;
+        [_reactSubviews insertObject:annotation atIndex:atIndex];
     }
 }
 
@@ -149,12 +148,13 @@
     // underlying mapview action here.
     if ([subview isKindOfClass:[RCTMapboxAnnotation class]]) {
         RCTMapboxAnnotation * annotation = (RCTMapboxAnnotation *) subview;
-        [_reactSubviews removeObjectForKey:annotation.reuseIdentifier];
+        [_map removeAnnotation:annotation];
+        [_reactSubviews removeObject:annotation];
     }
 }
 
 - (NSArray<id<RCTComponent>> *)reactSubviews {
-    return nil;
+    return _reactSubviews;
 }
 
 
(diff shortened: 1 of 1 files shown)


atarsha:
This solves a large number of issues that I had been having with custom annotations!

I am experiencing one thing that is a little strange, however. In using your branch temporarily, I'm finding that custom annotations actually cover my current position indicator on the map.

I'm not sure if this can be corrected on my end with some styling, or if this is a change I hadn't noticed before due to using a previous version of the Mapbox lib, but it is a change in behavior from what I had experienced previously.

Here is a screenshot of that behavior:
image

In addition, when testing on Android tonight, I found that it does not behave this way, HOWEVER, my custom annotations are TouchableHighlights, and they are not firing onPress unless I turn off showsUserLocation.

It's as if the user location layer is consuming the tap gesture and not passing it down to the custom annotations. On iOS, the annotations are on top, so they work fine.

dorthwein:
This fix solved the issue for us as well.  PLEASE MERGE ASAP!!! +1

atarsha:
@jackfengji Any ideas on my comment above? Do you think your PR introduced this? Any ideas on how to work around this?

jackfengji:
@atarsha 
1. I think this is the default behavior of my location annotation in iOS though I didn't try this in native mapbox-gl SDK. But my PR doesn't change this part at all.
2. The Android version of this RN SDK has big problems in touch event handling. Anyview can intercept the touch events now which is not good. I'm trying to solve the touch issue now but haven't got a perfect solution.

atarsha:
@jackfengji I didn't experience this behavior on iOS prior to switching to your fork. That said, it's hard for me to say that it is certainly due to your changes because of the random crashes associated with this bug prior to your fix. I can't help but wonder if your (I think it was 'atIndex') on the add layers method calls should be tweaked? I don't know if that's a good idea, I'm just guessing a bit here.

As for the Android touch event handling, it's probably outside the scope of this PR, but it seems to me like the showsUserLocation indicator should always pass touch events down.

markhaasjes:
+1

Jose4gg:
@jackfengji I having problems when I try to zoom out/in and move the map, the annotations get stuck on the screen and no matter what I do, even remove all annotations using JS, they don't go.

jackfengji:
@Jose4gg I spent a lot of time on the issue you mentioned then I fixed it. It's not related to my change. Please check https://github.com/waijule/react-native-mapbox-gl/commit/87a76becf9eaf3f803c2fa52ad8ae889bfe63f51 to fix stuck issue. I didn't submit a PR because I added many more things to the lib in my own branch. You can cherry-pick to your local branch to work with.

Jose4gg:
@jackfengji Man you are amazing!! Thanks for all that help, really appreciate it!!

atarsha:
@jackfengji I can't explain why, but if I change `return _reactSubviews;` back to `return nil;` on line 157, the annotations appear underneath the userLocation indicator as expected.

ios/RCTMapboxGL/RCTMapboxGL.m
```diff
@@ -149,12 +148,13 @@ - (void)removeReactSubview:(id<RCTComponent>)subview {
...
         RCTMapboxAnnotation * annotation = (RCTMapboxAnnotation *) subview;
-        [_reactSubviews removeObjectForKey:annotation.reuseIdentifier];
+        [_map removeAnnotation:annotation];
+        [_reactSubviews removeObject:annotation];
     }
 }
 
 - (NSArray<id<RCTComponent>> *)reactSubviews {
-    return nil;
+    return _reactSubviews;
```
atarsha, line 157: Changing this causes the annotation to render in front of the userLocation indicator. I tested switching this back to nil, and the annotation rendered behind the userLocation as expected. I don't know what implications changing this back has, but I think it should be looked in to. Otherwise this PR solves a huge issue for my team.

atarsha (changes requested):
Please review my inline comment.

atarsha:
@jackfengji just following up on my comment above. I don't exactly know what kind of impact changing that back would have so I would love for you to take a look and tell me if that's a feasible change to the PR? Thanks!

atarsha:
@zugaldia The issue this PR addresses causes a crash for nearly anyone using custom annotation views. Despite the fact that the contributor hasn't responded to my questions, the issue I'm seeing and awaiting a response about is **far** less severe than the crash is, which this solves.

Can we merge this, and if we don't hear an answer from @jackfengji, I'll investigate creating an additional PR afterwards to make the change I asked about above which seems to address the issue I'm seeing on iOS?

atarsha:
After messing with this further, I found that by changing that line back to nil like I suggested above, yes it makes it so that the location is on top as expected, but it then breaks the re-use of the view. Meaning, if you change it back to nil, removing the custom annotation does not remove the view from the screen.

I think this should be merged as-is! @zugaldia

atarsha (approved):

jjzazuet:
Hi. I'm also running into this issue. Any chances this PR can get merged into `master`? Thanks!

atarsha:
@nitaliano This PR also addresses a major pain point for many. If this could get reviewed and included in the next release it would be fantastic. Most of us have forked in order to merge this PR in.

Thanks!

henrikra:
I think we should ping @nitaliano to check this out ;)

nitaliano:
:+1:

henrikra:
@nitaliano Can you push 5.2.1 to npm since I can't install it with npm or yarn :)

PR #563, opened 2017-05-30
Android improvements

kormang:
Simplified android code responsible for adding and removing children to ReactNativeMapboxGLView. This should also improve performance, although there was no performance measuring.

Also there was a bug in getBounds method, which is now fixed in relatively acceptable way.


android/src/main/java/com/mapbox/reactnativemapboxgl/ReactNativeMapboxGLView.java
```diff
@@ -682,7 +726,13 @@ public CameraPosition getCameraPosition() {
     }
 
     public LatLngBounds getBounds() {
-        if (_map == null) { return new LatLngBounds.Builder().build(); }
+        if (_map == null) {
+            LatLng center = _initialCamera.build().target;
+            return new LatLngBounds.Builder()
+                    .include(center)
+                    .include(center)
```
henrikra: Why `.include(center)` is twice here?

kormang: Oh, probably left over after conflict resolution. I'll fix it.

kormang: Oh, it's because including center twice works, there is no problem with that. We need two points provided to builder, and it is not complaining if both are same. I can leave comment in code if it will make difference.

kormang: Without it this code will throw exception. I though this is acceptable way to fix it. At least it works for me, and it is certainly better that getting red screen with stack trace printed and having no way of controlling it.

henrikra: Yeah I think it should be commented then because it looks like a mistake. There is no other way of doing it?

kormang: Maybe there is, but I think it requires broader discussion about behavior of the whole system. Maybe we should prevent this method of even being called when _map is null. But as I said it might require broader discussion about it, and larger refactoring. As I'm using MapboxGL, I'm open for cooperation, and would like to make better pull request. Again, this is certainly better that it used to be, and in my project it works in acceptable way.

alfonsodev:
Hello, Thank you for this PR,
what is its status ?, I'm having the issue described in #586 and that this PR fixes. 
I see the branch has now conflicts, if you are willing to merge this I could help with
trying to fix the conflicts.
Thanks.

kormang:
Hello @alfonsodev . I have not received much attention and reviews, so basically I gave up. I suppose it's because all work was directed toward v6 API. I'm willing to resolve conflict, I don't think this commit is relevant to new API v6, and I don't know how soon it will be merged. I can also pick only does commit that you are willing to accept and make new pull request. Personally I think it all works well. :)

kormang:
Now I see what the conflict was, I was pretty sure, I had solved that same conflict locally...strange.

kormang:
I have committed new fix.

nitaliano:

Issue #593, opened 2017-07-25
mapbox/mapbox.h file not found


zayar:
I just install it but build fail because of 'Mapbox/Mapbox.h file not found in ios.

I tried to install both iOS (manually) and CocoaPods

henrikra:
I did fresh install with RN 0.46 and RNMapboxGL 5.2.0 and it worked. What else you tried?

alexaung:
I tried with 
"react-native": "0.46.3",
"react-native-mapbox-g": "5.2.0".

But got the same error for fresh install.

henrikra:
What have you done so far? Did you do all steps in manual way? Also did you rebuild the app (not just refresh) when everything was done? Be more specific :D

JClackett:
Did all the steps as per the documentation, and getting same error.

Am using the same versions of react-native/mapbox as alexaung

<img width="680" alt="screen shot 2017-07-31 at 16 02 30" src="https://user-images.githubusercontent.com/12549124/28781241-b67caa46-7609-11e7-9497-162eb3687d1f.png">

henrikra:
Please make fresh repo and try to install iOS manually and send link for it. This way we can see your problem :)

JClackett:
Managed to get it working, must have not installed correctly..

henrikra:
@JClackett Great please close the issue :)

danieljvdm:
This has been happening to me over the past few days. It works for a day and then randomly stops working and it's driving me nuts. I've tried manually linking as well as through coccoapods, and they all work for awhile until they stop working after changing nothing to do with the library or the Xcode project.

henrikra:
Looks like for some reason `Mapbox.framework` disappears  from node_modules so it gives error `mapbox/mapbox.h file not found`. Proper fix for this would be to always include `Mapbox.framework` like the PR https://github.com/mapbox/react-native-mapbox-gl/pull/595 is doing

henrikra:
@zayar @alexaung @danieljvdm Please check that your `Mapbox.framework` exists at `node_modules/react-native-mapbox-gl/ios/Mapbox.framework`. If it does not there is your problem :)

danieljvdm:
Yeah this is the issue. I think it's related to whenever you run `npm install` or `yarn install` it's refreshing the modules and deleting the framework for some reason, so I have to uninstall and reinstall every time.

studyroz:
[...]

PR #595, opened 2017-07-28
[Proposal] Add vendoring for Mapbox.framework to npm package

kitten:
During installation the entire `Mapbox.framework` is downloaded using `curl`. Unfortunately it fails quite easily on unstable network connections, and retrying to download starts over completely.

Instead this entire problem can be side-stepped by vendoring `Mapbox.framework`. It doesn't have to be added to the repo, as before, but we can add it to the npm package.

~Additionally this PR fixes the `podspec` file to point to the framework's headers, which was missing before.~ (Moved to #607)


henrikra:
I think this helps with https://github.com/mapbox/react-native-mapbox-gl/issues/593

.npmignore
```diff
@@ -37,3 +37,6 @@ android/gradle/
 android/gradlew
 android/gradlew.bat
 android/local.properties
+
+# Vendor iOS Mapbox.framework
+!ios/Mapbox.framework
```
henrikra, line 42: Why `!` is there?

package.json
```diff
@@ -22,7 +22,7 @@
     "url": "https://github.com/mapbox/react-native-mapbox-gl"
   },
   "scripts": {
-    "preinstall": "node ./scripts/download-mapbox-gl-native-ios-if-on-mac.js 3.3.1",
+    "preversion": "./scripts/download-mapbox-gl-native-ios.sh 3.3.1",
```
henrikra, line 25: What is different with `preinstall` and `preversion`?

.npmignore
kitten: This includes the framework in the tarball that will be sent to the npm registry

package.json
kitten: This runs the hook before a version is published (npm version) instead of when the user installs the package

henrikra: Ah okay! This is good!

henrikra: Also if the `download-mapbox-gl-native-ios.sh` fails it won't publish new version?

kitten: @henrikra I haven't checked whether it actually aborts 🤔  I'll have to check that

danieljvdm:
What is the status on this PR? I think it's pretty necessary to fix a lot of linking errors.

kitten:
@danieljvdm afaik this repo is currently in a small break of maintenance, but I'll see if I can contact some people in the meantime

nitaliano:
@philpl This is a good idea! We can work this into the repo once we start it's refactor.  For now I think we should hold off on merging this in and just merge in the podspec portion.

kitten:
@nitaliano if you’d like I can move the pod spec fixes into a separate PR?

nitaliano:
yes that would be great if you can

kitten:
@nitaliano :smile: This PR doesn't include the podspec modifications anymore. Those are now in a separate PR here: https://github.com/mapbox/react-native-mapbox-gl/pull/607

henrikra:
@philpl There is still one conflict

PR #599, opened 2017-08-04
Allow user to tap on annotation callout windows

nolan-m:
- Added ability to tap on an annotation callout window.
- Removed `onRightAnnotationTapped` from firing on Android since callout accessories are not supported. 
- Updated example and docs.


nitaliano:

PR #601, opened 2017-08-07
Make offline packs more robust.

gillessed:
Some small changes to make the library a little more robust when used.

- Separate the access token initialization and offline maps
  initialization.
- Make the initialization functions return promise so that the consumer
  code can know when mapbox and the offline packs are ready to be used, instead of just blindly 
  hoping things will work.
- Expose setConnected method.
- Update docs.


gillessed:
I am going to make the changes for ios shortly.

PR #603, opened 2017-08-09
Fix ios offline pack bugs.

gillessed:
- Only resume packs that aren't complete.
- Don't fire events for invalid packs.


gillessed:
@zugaldia I have some more small fixes.

zugaldia (approved):

PR #613, opened 2017-08-22
Added Typescript type definitions. Closes #608

### LINKED ISSUE #608: Add typescript definition types
Can we add typescript support by adding ts definition types directly to this repo? I wrote them, and I would prefer to have them in this repo, rather than on definitelyTyped.

madroneropaulo:
Added type definitions for Typescript. See #608


henrikra:
Can you reopen this @nitaliano ? We need TS types for this project! :)

PR #619, opened 2017-09-01
MVP Version 6

nitaliano:
Just creating a PR to allow making code reviews easier, I don't plan on merging this in. Would love to get some feedback on the code so far.

**cc** @bsudekum @1ec5 @boundsj @zugaldia @tobrun


example/src/App.js
```diff
@@ -0,0 +1,172 @@
...
+
+    this.renderItem = this.renderItem.bind(this);
+    this.onCloseExample = this.onCloseExample.bind(this);
+  }
+
+  async componentWillMount () {
+    if (IS_ANDROID) {
+      await MapboxGL.requestPermissions();
+    }
+    MapboxGL.setAccessToken('pk.eyJ1Ijoibmlja2l0YWxpYW5vIiwiYSI6ImNqNjVtemx3dzAyeHkyd3IwOHJ3b3JsNmMifQ.QSvBB90og-7PW68dZPDAIg');
```
bsudekum: Don't commit access tokens.

nitaliano: Good catch! Updated

example/src/App.js
```diff
@@ -0,0 +1,173 @@
...
+
+    this.renderItem = this.renderItem.bind(this);
+    this.onCloseExample = this.onCloseExample.bind(this);
+  }
+
+  async componentWillMount () {
+    if (IS_ANDROID) {
+      await MapboxGL.requestPermissions();
+    }
+    MapboxGL.setAccessToken(config.get('accessToken'));
```
bsudekum: I don't think we should be dictating to developers how they should be configuring private/config variables. I think inserting the string `INSERT MAPBOX ACCESS TOKEN HERE. Your access token can be found here: Link` should be enough.

example/src/components/SetUserTrackingModes.js
```diff
@@ -0,0 +1,49 @@
...
+    ...BaseExamplePropTypes,
+  };
+
+  constructor (props) {
+    super(props);
+
+    this._trackingOptions = Object.keys(MapboxGL.UserTrackingModes).map((key) => {
+      return {
+        label: key,
+        data: MapboxGL.UserTrackingModes[key],
```
bsudekum, line 21: `trackingMode` might be more descriptive here.

ios/RCTMGL/MGLModule.m
```diff
@@ -0,0 +1,78 @@
...
+
+- (NSDictionary<NSString *, id> *)constantsToExport
+{
+    // style urls
+    NSMutableDictionary *styleURLS = [[NSMutableDictionary alloc] init];
+    [styleURLS setObject:[MGLStyle.streetsStyleURL absoluteString] forKey:@"Street"];
+    [styleURLS setObject:[MGLStyle.darkStyleURL absoluteString] forKey:@"Dark"];
+    [styleURLS setObject:[MGLStyle.lightStyleURL absoluteString] forKey:@"Light"];
+    [styleURLS setObject:[MGLStyle.outdoorsStyleURL absoluteString] forKey:@"Outdoors"];
+    [styleURLS setObject:[MGLStyle.satelliteStyleURL absoluteString] forKey:@"Satellite"];
```
bsudekum, line 33: There are a few more styles missing from this list: https://www.mapbox.com/api-documentation/#styles

ios/RCTMGL/MGLModule.m
```diff
@@ -0,0 +1,78 @@
...
+    [eventTypes setObject:RCT_MAPBOX_DID_FINISH_LOADING_STYLE forKey:@"DidFinishLoadingStyle"];
+    [eventTypes setObject:RCT_MAPBOX_FLY_TO_COMPLETE forKey:@"FlyToComplete"];
+    [eventTypes setObject:RCT_MAPBOX_SET_CAMERA_COMPLETE forKey:@"SetCameraComplete"];
+    
+    // user tracking modes
+    NSMutableDictionary *userTrackingModes = [[NSMutableDictionary alloc] init];
+    [userTrackingModes setObject:[NSNumber numberWithInt:MGLUserTrackingModeNone] forKey:@"None"];
+    [userTrackingModes setObject:[NSNumber numberWithInt:MGLUserTrackingModeFollow] forKey:@"Tracking"];
+    [userTrackingModes setObject:[NSNumber numberWithInt:MGLUserTrackingModeFollowWithHeading] forKey:@"Compass"];
+    [userTrackingModes setObject:[NSNumber numberWithInt:MGLUserTrackingModeFollowWithCourse] forKey:@"Navigation"];
```
bsudekum: I might follow the tracking mode names more closely here. `.none`, `.follow`, `.followWithHeading`, `.followWithCourse`.

ios/RCTMGL/RCTMGLMapViewManager.m
```diff
@@ -0,0 +1,287 @@
...
+
+@interface RCTMGLMapViewManager() <MGLMapViewDelegate>
+@end
+
+
+@implementation RCTMGLMapViewManager
+
+// prevents SDK from crashing and cluttering logs
+// since we don't have access to the frame right away
+static CGRect const RCT_MAPBOX_MIN_MAP_FRAME = { { 0.0f, 0.0f }, { 64.0f, 64.0f } };
```
bsudekum, line 28: Would it be possible to make this equal to the height/width of the parent view?

nitaliano: We don't have access to that information unfortunately when we initialize the view.  It hasn't been inserted into the view hierarchy at this point. RN behind the scenes on iOS will mess around with this view and set the frame based off it's parent and `style` attribute passed down from the JS later on once it has access to the information.

example/src/App.js
nitaliano: My only concern having it as a hardcoded string would be that every time someone wants to add an example they would have to add the token here and remove the token here before making a commit.

ios/RCTMGL/MGLModule.m
nitaliano: This is where things will get interesting, the iOS and Android SDKs have different names for the modes. Unless I am looking in the wrong place.

https://github.com/mapbox/mapbox-plugins-android/blob/master/plugins/locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerMode.java

atarsha: I ran into similar issues while working on a PR you recently merged. Android and iOS have different OfflinePackStates, so I had to create a method to normalize them into one. In that particular case, iOS had more states than Android, so it made sense to map them to the iOS states. In this particular case, I agree with @bsudekum. The iOS names are more descriptive. The actual result here isn't "compass" it's "follow the user with compass rotation." I also find "Tracking" vs "Compass" not very self-descriptive. While I know the answer to these questions, here is an example: If following the user is "tracking" then is "compass" not following the user and just rotating?

Ideally, this would be fixed upstream, but since it isn't, I think we should run with more descriptive, even if it means choosing our own names entirely for RN.

android/rctmgl/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.mapbox.reactnativemapboxgl" >
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="mapbox.rctmgl">
```
jackfengji: Why do you use mapbox.rctmgl instead of com.mapbox.rctmgl?

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
```diff
@@ -0,0 +1,180 @@
...
+
+    //region Custom Events
+
+    @Override
+    public Map<String, String> customEvents() {
+        return MapBuilder.<String, String>builder()
+                .put(EventKeys.MAP_CLICK, "onPress")
+                .put(EventKeys.MAP_LONG_CLICK,"onLongPress")
+                .put(EventKeys.MAP_ONCHANGE, "onMapChange")
+                .put(EventKeys.USER_LOCATION_CHANGE, "onUserLocationChange")
```
jackfengji: I don't think we should have this event in mapbox SDK. User should listen to location changes from native SDKs.

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -0,0 +1,438 @@
...
+        mMap.easeCamera(update, durationMS);
+        mMap.easeCamera(update, durationMS, new SimpleEventCallback(mManager, event));
+    }
+
+    public void fitBounds(FeatureCollection featureCollection, int padding, int durationMS) {
+        LatLngBounds bounds = ConvertUtils.toLatLngBounds(featureCollection);
+        if (bounds == null) {
+            return;
+        }
+        mMap.easeCamera(CameraUpdateFactory.newLatLngBounds(bounds, padding), durationMS);
```
jackfengji: Do we need to handle the situation that user calls fitBounds/flyTo before map ready? For example in componentDidMount

android/rctmgl/src/main/AndroidManifest.xml
nitaliano: It should be `com.mapbox.rctmgl` good catch

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
nitaliano: Yes, we absolutely should.

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
nitaliano: I'm all for dropping this the less updates going over the bridge from this repo the better.

android/rctmgl/src/main/AndroidManifest.xml
nitaliano: Just pushed a commit to fix the package

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
nitaliano: Updated the PR, I put all camera updates on a queue now so something like this is now possible

```
this.map.setCamera({
  stops: [
    { pitch: 45, duration: 200 },
    { centerCoordinates: [...], zoom: 16, mode: MapboxGL.CameraModes.Flight },
    { heading: 180, duration: 2000 },
    { heading: 360, duration: 2000 },
    {
       bounds: {
          ne: [...],
          sw: [...],
          padding: 15
       }
    }
  ]
})
```

ios/RCTMGL/MGLModule.m
nitaliano: Updated

ios/RCTMGL/MGLModule.m
nitaliano: I'll switch the names over. I agree we should be using more descriptive names.

atarsha:
I created a wiki page that anyone should feel free to add to with a few suggestions for you for V6.

https://github.com/mapbox/react-native-mapbox-gl/wiki/V6-Suggestions

If this would be better suited as an issue, I'm happy to convert it.

android/rctmgl/src/main/java/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
atarsha: Just a reminder forget to remove that event from docs/MapView.md as well.

EDIT: Disregard, I see that is a result of auto-generating docs, and will be removed automatically once removed from the code.

1ec5:
Thanks for your suggestions @atarsha! It might be good to get some of the more actionable ones into dedicated issues so everyone can track the progress on them more closely. Off the top of my head, the suggestion about shape selection would be blocked by mapbox/mapbox-gl-native#2082, although it’s also dependent on whether this library continues to support shape annotations versus implementing them using runtime styling.

ios/RCTMGL/RCTMGLMapView.m
```diff
@@ -78,6 +93,30 @@ - (void)setReactUserTrackingMode:(int)reactUserTrackingMode
...
+    double constrainedZoom = [[RCTMGLUtils clamp:[NSNumber numberWithDouble:zoomLevel]
+                                             min:[NSNumber numberWithDouble:self.minimumZoomLevel]
+                                             max:[NSNumber numberWithDouble:self.maximumZoomLevel]] doubleValue];
+    
+    double constrainedLatitude = [[RCTMGLUtils clamp:[NSNumber numberWithDouble:latitude]
+                                                 min:[NSNumber numberWithDouble:-LAT_MAX]
+                                                 max:[NSNumber numberWithDouble:LAT_MAX]] doubleValue];
+    
+    double constrainedScale = pow(2.0, constrainedZoom);
+    return cos(constrainedLatitude * DEG2RAD) * M2PI * EARTH_RADIUS_M / (constrainedScale * TILE_SIZE);
```
1ec5, line 239: I suspect you could take the result of `-[MGLMapView metersPerPointAtLatitude:]` and scale it by the zoom level without having to perform the above math independently.

ios/RCTMGL/RCTMGLMapView.m
```diff
@@ -78,6 +93,30 @@ - (void)setReactUserTrackingMode:(int)reactUserTrackingMode
     [self setUserTrackingMode:(NSUInteger)_reactUserTrackingMode animated:_animated];
 }
 
+#pragma mark - methods
+
+- (CLLocationDistance)getMetersPerPixelAtLatitude:(double)latitude withZoom:(double)zoomLevel
```
1ec5: The usual case would be to convert from pixels to meters at the current zoom level. Maybe make the zoom level optional?

ios/RCTMGL/RCTMGLMapView.m
```diff
@@ -78,6 +93,30 @@ - (void)setReactUserTrackingMode:(int)reactUserTrackingMode
...
+    double constrainedScale = pow(2.0, constrainedZoom);
+    return cos(constrainedLatitude * DEG2RAD) * M2PI * EARTH_RADIUS_M / (constrainedScale * TILE_SIZE);
+}
+
+- (CLLocationDistance)altitudeFromZoom:(double)zoomLevel
+{
+    CLLocationDistance metersPerPixel = [self getMetersPerPixelAtLatitude:self.camera.centerCoordinate.latitude withZoom:zoomLevel];
+    CLLocationDistance metersTall = metersPerPixel * self.frame.size.height;
+    CLLocationDistance altitude = metersTall / 2 / tan(MGLRadiansFromDegrees(30) / 2.0);
+    return altitude * sin(M_PI_2 - MGLRadiansFromDegrees(self.camera.pitch)) / sin(M_PI_2);
```
1ec5, line 247: mapbox/mapbox-gl-native#5583 would be nice to have.

bsudekum:
@nitaliano would it make sense to publish an v6 alpha to allow developers to get a feel of how the new APIs will work?

nitaliano:
@bsudekum Yes, I'm updating the doc generating to handle methods and I'll update the README and we can publish an alpha

ios/RCTMGL/RCTMGLMapView.m
nitaliano: Yes it would!

nitaliano:
@atarsha awesome! I think it could be good to have a few issues for discussion.

* Runtime styling (First PR happening end of this week early next week, could be good to discuss in the PR)
* Annotations
* Offline (I'm planning on porting over what we have)

These are the big features I think we need to ship with v6

nitaliano:
Just wanted to update this PR, now that all of the features have been implemented I am planning on merging this into master on Friday.

Any issues with that let me know.

hannta:
Im able to reproduce this gps issue #692 on android with `6.0.0-beta1` and I do not see any recent commits which should fix this?

nitaliano:
Going to hold off merging this in until Monday

nitaliano:
It's happening

PR #626, opened 2017-09-09
setVisibleCoordinateBounds accepts object rather than 9 parameters

henrikra:
It is very confusing to call `setVisibleCoordinateBounds` since it has 9 parameters o.O
Example:
```js
this.mapView.setVisibleCoordinateBounds(
  firstMarker.lat,
  firstMarker.lon,
  secondMarker.lat,
  secondMarker.lon,
  50,
  15,
  0,
  15,
);
```
What the hell this means? :D It means that you want padding 50 to top, padding 15 to left and right and also padding 0 to bottom.

Instead calling like this makes much more sense and increases readability
```js
this.mapView.setVisibleCoordinateBounds({
  latitudeSW: firstMarker.lat,
  longitudeSW: firstMarker.lon,
  latitudeNE: secondMarker.lat,
  longitudeNE: secondMarker.lon,
  paddingTop: 50,
  paddingLeft: 15,
  paddingRight: 15,
});
```
Notice also that I did not have to add `paddingBottom` because I did not want to change it!

Also note that this is breaking change ⚠️

### DIFF
--- a/API.md
+++ b/API.md
@@ -115,7 +115,17 @@ animation has ended.
 ---
 
 ```javascript
-this._map.setVisibleCoordinateBounds(latitudeSW, longitudeSW, latitudeNE, longitudeNE, paddingTop = 0, paddingRight = 0, paddingBottom = 0, paddingLeft = 0, animated = true);
+this._map.setVisibleCoordinateBounds({
+  latitudeSW,
+  longitudeSW,
+  latitudeNE,
+  longitudeNE,
+  paddingTop = 0,
+  paddingRight = 0,
+  paddingBottom = 0,
+  paddingLeft = 0,
+  animated = true
+});
 ```
 
 This method adjusts the center location and the zoomLevel of the map so that
--- a/index.js
+++ b/index.js
@@ -237,8 +237,29 @@ class MapView extends Component {
     return promise;
   }
 
-  setVisibleCoordinateBounds(latitudeSW, longitudeSW, latitudeNE, longitudeNE, paddingTop = 0, paddingRight = 0, paddingBottom = 0, paddingLeft = 0, animated = true) {
-    MapboxGLManager.setVisibleCoordinateBounds(findNodeHandle(this), latitudeSW, longitudeSW, latitudeNE, longitudeNE, paddingTop, paddingRight, paddingBottom, paddingLeft, animated);
+  setVisibleCoordinateBounds({
+    latitudeSW,
+    longitudeSW,
+    latitudeNE,
+    longitudeNE,
+    paddingTop = 0,
+    paddingRight = 0,
+    paddingBottom = 0,
+    paddingLeft = 0,
+    animated = true,
+  }) {
+    MapboxGLManager.setVisibleCoordinateBounds(
+      findNodeHandle(this),
+      latitudeSW,
+      longitudeSW,
+      latitudeNE,
+      longitudeNE,
+      paddingTop,
+      paddingRight,
+      paddingBottom,
+      paddingLeft,
+      animated,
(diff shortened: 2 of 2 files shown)


franky47:
While it does indeed increate readability when calling it directly with explicit arguments, it's inconvenient to use the object parameter if the data one's working with is already formatted the right way.

Example:
```js
// before PR
const bounds = turf.bbox(someFeatureOrSetOfFeatures)
this.map.setVisibleCoordinateBounds(...bounds)

// after PR
const bounds = turf.bbox(someFeatureOrSetOfFeatures)
this.map.setVisibleCoordinateBounds({
  latitudeSW: bounds[0],
  longitudeSW: bounds[1],
  latitudeNE: bounds[2],
  longitudeNE: bounds[3]
})
```

henrikra:
Hmm can you give more full featured example of how you use `turf`? Also how you can set padding in the first example?

franky47:
Sure, [turf](http://turfjs.org/docs/) is an awesome companion to mapbox for GeoJSON manipulation:
```js
import Polyline from '@mapbox/polyline'
import turf from '@turf/turf'

const polyline = 'akncFlxghVxs@ut@br@yeApu@srAkNs}BgvBg{Amp@l{ByEvnCj|@mTrLen@'

// convert the encoded polyline into a GeoJSON LineString
const line = turf.lineString(Polyline.decode(polyline))

// fit the map around the bounds of that line
this.map.setVisibleCoordinateBounds(...turf.bbox(line))
```

henrikra:
Cool! How you can set padding in this line btw? `this.map.setVisibleCoordinateBounds(...turf.bbox(line))`

franky47:
Either specify them after the bounds, or use more array de-structuring  sugar:
```js
const padding = [30, 30, 30, 30]
this.map.setVisibleCoordinateBounds(...turf.bbox(line), ...padding)

// Fixed padding is even easier with ES6
const padding = 30
this.map.setVisibleCoordinateBounds(...turf.bbox(line), ...Array(4).fill(padding))
```
Now since padding in Mapbox is expressed in screen points, sometimes you might want to express it in relation to the size of the feature to fit. Turf allows that:
```js
this.map.setVisibleCoordinateBounds(...turf.bbox(turf.transformScale(line, 1.25))
```

henrikra:
What about doing like this?
```js
const [latitudeSW, longitudeSW, latitudeNE, longitudeNE] = turf.bbox(someFeatureOrSetOfFeatures)
this.map.setVisibleCoordinateBounds({latitudeSW, longitudeSW, latitudeNE, longitudeNE})
```

franky47:
The issue here is no longer remembering the order of the arguments in `setVisibleCoordinateBounds`, but both their order in what `turf.bbox` returns and their name in the object argument.

henrikra:
Hmm you can name `const [latitudeSW, longitudeSW, latitudeNE, longitudeNE]` what ever you want

franky47:
Except that the parameter object expects these names. That seems like an unnecessary indirection.

nitaliano:
I would be a fan of

```
ne = [lng, lat]
sw = [lng, lat]

setVisibleCoordinateBounds(ne, sw, [top, right, bottom, left])
setVisibleCoordinateBounds(ne, sw, [vertical, horizontal])
setVisibleCoordinateBounds(ne, sw, [all]) or setVisibleCoordinateBounds(ne, sw, all)
```
makes it more css like

henrikra:
I will close this then :P

ansarikhurshid786:
I am getting error : 
setVisibleCoordinateBounds is not a function

PR #650, opened 2017-09-28
Update Xcode project file name

stdavis:


stdavis:
I didn't realize that I was working with v6 and it looks like the iOS doc for the `v6` branch is correct making this PR irrelevant.

PR #654, opened 2017-10-03
[Android] Preserve width and height for marker from parameters

keenubee:


nitaliano:

PR #659, opened 2017-10-05
Minor tweaks

jliebrand:
Tripped me up, so thought I'd update the readme...


jliebrand:
Also, more of a question, but the example app does not have the bash script in the Build Phases as outlined by the readme; is it still needed?

nitaliano:
@jliebrand the scripts not needed to get anything working, so for an example you wouldn't need it. For a production app you would want to include the script, because it removes some stuff.

Thanks for the contribution!

jliebrand:
Actually - just realised, the npm instructions should probably just be:
npm install --save @mapbox/react-native-mapbox-gl

without the version number - since the version in the readme is already now outdated

PR #666, opened 2017-10-14
Better version tracking on Mapbox.framework downloads

kevinsperrine:
Added a `.framework_version` file to check against and only delete and re-download the Mapbox.framework when the version has changed.

### DIFF
--- a/scripts/download-mapbox-gl-native-ios.sh
+++ b/scripts/download-mapbox-gl-native-ios.sh
@@ -1,9 +1,14 @@
 #!/bin/sh
 
-VERSION=$1
-
-echo "Downloading Mapbox GL iOS $VERSION, this may take a minute."
 cd ios/
+
+VERSION=$1
+CURRENT_VERSION=$(cat .framework_version)
+
+if [ "$VERSION" == "$CURRENT_VERSION" ]; then
+  echo "The newest version is already installed. Exiting."
+  exit 0
+fi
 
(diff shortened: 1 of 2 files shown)


scripts/download-mapbox-gl-native-ios.sh
```diff
@@ -1,9 +1,16 @@
 #!/bin/sh
 
+cd ios/
+
 VERSION=$1
+CURRENT_VERSION=$(cat .framework_version)
```
nitaliano, line 6: I like the flexibility this gives with being able to update the version independently. I do think we should also provide a default version number when `.framework_version` does not exist, for a few reasons

- Know the version of the Mapbox.framework that this lib was tested to work with.
- Allow users with not much experience in the mobile space to just be able to get there projects up and running without needing to add this file.

kevinsperrine: For sure, I left that part in package.json the way it's currently done. This change was purely to prevent Mapbox.framework from being deleted/redownloaded each time there is a dependency change if the Mapbox verison wasn't changed. I added it because I was traveling this weekend and used a hotspot a lot, so deleting/redownloading ~60MB each time I added a new dependency got annoying. :-)

nitaliano: ha, I misread it cool thanks for the PR!

Issue #671, opened 2017-10-17
[v6] SymbolLayer with style property textOffset


karlguillotte:
Hi,

Trying to use the `textOffset` on a `SymbolLayer` and it does not work. Below is a screenshot of the error.

![image](https://user-images.githubusercontent.com/744011/31686988-c9576b4a-b33c-11e7-8838-59597ace6a32.png)

Here is my style object:

```
const styles = Mapbox.StyleSheet.create({
    clusterLabel: {
        textSize:12 ,
        textColor: BLACK,
        textHaloColor: WHITE,
        textHaloWidth: 1,
        textField: '{point_count}',
        textOffset: [2, 2],
    }
})
```

I wanted to test with the demo application, but I ran into issues with a `babel-preset-react-native` transformation trying to start it. 

Thanks!

nitaliano:
Theres's a bug with the version of `babel-preset-react-native` in the demo app, if you switch to node 8 it will run.

I'll look into textOffset and probably the rest of the SymbolLayer styles to make sure they're working correctly.

nitaliano:
I already see the problem. Some offsets require `NSValue` that is a `CGVector` others require `NSArray<NSNumber *> *`.

karlguillotte:
Thanks @nitaliano ! Ready to test as soon as you have a patch.

nitaliano:
@karlguillotte I added the fix for this on the PR that fixes the iconImage token not working. https://github.com/mapbox/react-native-mapbox-gl/pull/674. Let me know how it goes.

karlguillotte:
It is working! Thanks @nitaliano.

nitaliano:
The fix has been merged into the v6 branch closing out

Issue #668, opened 2017-10-16
[v6] SymbolLayer with style { iconName:{token} }


karlguillotte:
Thanks for your work on the v6 branch!

I am trying to add a Mapbox.SymbolLayer with a style with an iconName value as a token. All other ways work (commented bellow)!

```js
const styles = Mapbox.StyleSheet.create({
    reports: {
        // iconImage: 'min-pin-with-incident',
        // iconImage: 'min-pin',
        // iconImage: require('../assets/min-pin-with-incident'),
        iconImage: '{icon}',
    },
})
```
And in my component's render method:
```js
...
render() {
  return (
    <Mapbox.MapView>
      <Mapbox.ShapeSource id="reports" shape={sources.reports}>
        <Mapbox.SymbolLayer id="reports" style={styles.reports} />
      </Mapbox.ShapeSource>
    </Mapbox.MapView>
  )
}
...
```
Here is the data for my Mapbox.ShapeSource (sources.reports):
```json
{"type":"FeatureCollection","features":[{"type":"Feature","id":"9d10456e-bdda-4aa9-9269-04c1667d4552","properties":{"icon":"min-pin"},"geometry":{"type":"Point","coordinates":[-117.20611157485109,52.18096108426141]}}]}
```

Not sure if it has been tested and implemented yet, but it seems it is when reading the docs https://github.com/mapbox/react-native-mapbox-gl/blob/v6/docs/SymbolLayer.md. 

Here is the warning I am getting. Looks like the token is not parsed and replaced. 
```
Could not find image file:///Users/karlguillotte/Library/Developer/CoreSimulator/Devices/C534294E-62BB-47EC-BFBA-FAF5A60C27BE/data/Containers/Bundle/Application/15739B69-797E-42D1-AE98-4088E74D3F5F/AvCanMobile.app/%7Bicon%7D.png

**Comment by nitaliano** [comment]
@karlguillotte thanks for posting this, the docs are auto generated from the [json style spec](https://github.com/mapbox/react-native-mapbox-gl/blob/v6/style-spec/v8.json), so they might be slightly ahead. I'll look into it and get it working.

**Comment by nitaliano** [comment]
Just wanted to update this ticket. I have a working implementation on iOS I just need to do the Android implementation then the PR will land.

The API I have in mind is below the `ShapeSource` will now be responsible for declaring any external images that we want to add to our geojson source, this will also ensure that the images are being cleaned up out of memory when the source is no longer in use.

```
[...]

PR #674, opened 2017-10-18
General Bugfixes

nitaliano:
Fixes https://github.com/mapbox/react-native-mapbox-gl/issues/671, https://github.com/mapbox/react-native-mapbox-gl/issues/668, https://github.com/mapbox/react-native-mapbox-gl/issues/679
* Adds support for iconImage string tokens.
* Fixes textOffset and iconOffset not using a `CGVector`.
* Adds example iconImage string tokens.
* Adds ability to create overrides for style types
* Adds missing `sourceLayerID` to `FillLayer`
* Adds example using custom vector source.


nitaliano:
@mapbox/ios would someone be able to review this https://github.com/mapbox/react-native-mapbox-gl/pull/674/files#diff-a70ee91d2efaddbcb649ad8bbc4f4ba0R66 for me and let me know if this is a stupid way of doing it.

pingnattawoot:
i think...
Should iconImage in SymbolLayer style come from sprite file ? 
just like mapbox-gl

boundsj:
> would someone be able to review this

This approach seems like it will work. Have you thought about edge cases like downloads taking a long time or failing or if you want to be able to cancel pending downloads? Also, is `fetchImages` blocking the main thread while it waits to signal?

It'd be more involved but you may want to consider an implementation using `NSOperationQueue` (a famous iOS image downloading implementation that uses that is https://github.com/rs/SDWebImage/blob/master/SDWebImage/SDWebImageDownloader.m). I can imagine scheduling the image downloads on an operation queue with custom operations that decrement your reference count and call the callback somehow when the count reaches 0.

For what you are trying to do here though, what you have may be fine. I'm not sure I fully understand how native code will be exercised (threading wise) in the context of RN!

nitaliano:
@pingnattawoot it should already support images coming from the sprite sheet and it should be able to handle external images as well that we have to register with the map style. So it would support
* Images from the JS layer
* Images from the native projects themselves(drawables, and *.xcassets)
* Images from the sprite sheet.

Issue #661, opened 2017-10-06
[Android] When showsUserLocation is true custom annotations don't receive touch event


henrikra:
If you set `showsUserLocation` to `true` on `MapView` and use custom markers inside of `MapView` the marker can't receive any touch events on Android. Therefore adding `TouchableOpacity` etc is useless. 

Hot fix for this is to set `showsUserLocation={false}` which is not ideal. 

React Native: 0.48.4
react-native-mapbox-gl: ^5.2.3-alpha1

Example code:
```js
<MapView 
  showsUserLocation={true}
>
  <Annotation
    id="badAssMarker"
    coordinate={...}
  >
    <TouchableOpacity onPress={() => this.showTimePicker('departure')}>
      <View style={{width: 75, height: 75, backgroundColor: 'red'}}>
        <Text >
          Hello there
        </Text>
      </View>
    </TouchableOpacity>
  </Annotation>
</MapView>
```

nitaliano:
Yes unfortunately this is a problem with the current marker implementation.  I'm working on the `v6` implementation right now and this is what the API will look like. Instead of having the custom view add the onPress the native code will handle this for us similar to how it's handled in react-native-maps.

```
<MapboxGL.PointAnnotation 
      id='annotation1'
      onPress={() => console.log('YAYY!')}
      style={{ width: 65, height: 65 }}
      coordinate={[lng, lat]}>

     <View style={....} />

</MapboxGL.PointAnnotation>
```

nitaliano:
Closing this out as we just merged in our `v6` rewrite. If this exists in the new version feel free to reopen.

Issue #647, opened 2017-09-26
[iOS] Touch does not go through custom marker on 2nd render


henrikra:
Gif of the problem is here: https://imgur.com/a/GoHzi
Reproduction repo here: https://github.com/henrikra/mapbox-custom-annotations

When I render custom marker without condition statement aka markers are rendered once then my touch events go through the marker which is good thing. But when my condition is true so that the second marker is rendered to the map my touches does not go through the markers at all! This is preventing me from dragging the map around which is very annoying.

In example code when `toMarker` is rendered both markers prevent touch event through them!

```js
<MapView>
  <Annotation
    id="fromMarker"
    coordinate={{
      latitude: fromLocation.location.lat,
      longitude: fromLocation.location.lon,
    }}
  >
    {/* stuff */}
  </Annotation>
  {this.state.toLocation && (
    <Annotation
      id="toMarker"
      coordinate={{
        latitude: toLocation.location.lat,
        longitude: toLocation.location.lon,
      }}
    >
      {/* stuff */}
    </Annotation>
  )}
</MapView>
```

Tell me if you need more info @nitaliano

henrikra:
@1ec5 @nitaliano @bsudekum I have now added GIF and reproduction repo so the bug is very clear:) Tell me guys if you need more info ;)

henrikra:
@jackfengji Can you check this out since you have touched the ios annotations the last :)?

martsie:
This is a similar issue to https://github.com/mapbox/react-native-mapbox-gl/issues/575 and https://github.com/mapbox/react-native-mapbox-gl/issues/559 where touches aren't being propagated to the mapview at all.

henrikra:
I think @nitaliano fixed these problems in v6 :)

nitaliano:
Closing this out as we just merged in our `v6` rewrite. If this exists in the new version feel free to reopen.

PR #677, opened 2017-10-19
[v6] Point Annotation Support

nitaliano:
* Adds support for PointAnnotations with custom views.
* Adds support for Callouts with custom views.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ npm install @mapbox/react-native-mapbox-gl@6.0.0-alpha1 --save
 * [MapView](/docs/MapView.md)
 * [Light](/docs/Light.md)
 * [StyleSheet](/docs/StyleSheet.md)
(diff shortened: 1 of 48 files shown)


henrikra:
Will this address problems on v5 also? #647 #661 ? So that they are fixed in v6?

nitaliano:
@henrikra I will test those issues to make sure, but I believe they won't be an issue. Are there any other important issues you know of that you think we should test against before merging?

henrikra:
@nitaliano 
1. There is sometimes issue also on Android that custom markers won't show on map until you touch the map :/
2. On Android touch will never get through custom markers. I think by default you can touch through custom marker and if you want them to respond to touch then user adds TouchableOpacity etc

Those came to my mind :)

nitaliano:
@henrikra I attempted to reproduce some of those issues today and I could not. Keeping my fingers crossed that those issues are resolved with this rewrite.

PR #689, opened 2017-10-25
[v6] Icon sprite sheet fix

nitaliano:
* Fixes support for custom sprites coming from the sprite sheet.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/RCTMGLStyle.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/RCTMGLStyle.java
@@ -1,8 +1,10 @@
 package com.mapbox.rctmgl.components.styles;
 
+import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.support.annotation.StringDef;
 
+import com.facebook.common.util.UriUtil;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.ReadableMapKeySetIterator;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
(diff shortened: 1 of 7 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/RCTMGLStyle.java
```diff
@@ -57,16 +59,24 @@ public RCTMGLStyleValue getStyleValueForKey(String styleKey) {
...
             return;
         }
-
         Map.Entry[] images = new Map.Entry[]{ new AbstractMap.SimpleEntry(uriStr, uriStr) };
         DownloadMapImageTask task = new DownloadMapImageTask(mMap, null);
         task.execute(images);
     }
 
+    private boolean shouldAddImage(String uriStr) {
+        return uriStr != null && !isTokenString(uriStr) && isValidURI(uriStr);
```
terryttsai: Do we still need `!isTokenString` when we already have `isValidURI`?

ios/RCTMGL/RCTMGLStyle.m
```diff
@@ -1847,6 +1847,11 @@ - (void)setIntensityTransition:(MGLLight *)layer withReactStyleValue:(RCTMGLStyl
 
 
 
+- (BOOL)_shouldAddImage:(NSString *)str
+{
+    return str != nil && ![self _isTokenString:str] && [self _isValidURL:str];
```
terryttsai: Same comment: Do we still need `!isTokenString` when we already have `isValidURI`?

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/RCTMGLStyle.java
nitaliano: Good catch, we do not I'll update it.

terryttsai:
The fix works for me on ios and android! 👍

PR #705, opened 2017-11-01
[WIP] [V6] Improvements

nitaliano:
* Fixes typo in Android autogen script.
* Updates Base layer, and source classes to use an existing layer/source if it exists based on id and not just composite
* Updates MapboxStyleSheet to be compatible with Animated values.
* Update all Layer.style to handle arrays just like View.style
* Add Animated wrapper for Layers and ShapeSource.

out


henrikra:
Wow this animating part is good news 🎂

PR #738, opened 2017-11-11
[v6] Cocoapods

### LINKED ISSUE #733: Add CocoaPods Support
We want to support CocoaPods, so we need to add a working podspec to this repo.

jesseditson:
**changes:**

- Update podfile to work with latest iOS SDK
- Add instructions for installing with CocoaPods

**summary:**

Resolves #733 - Looks like cocapods support was mostly there (it was working in v5.x), just needed the Mapbox SDK dependency updated (and instructions).

### DIFF
--- a/ios/install.md
+++ b/ios/install.md
@@ -1,5 +1,18 @@
 # iOS Installation
 
+## Using CocoaPods
+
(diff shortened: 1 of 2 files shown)


nitaliano:
Thanks for the PR!

Issue #736, opened 2017-11-10
[v6] Application crashes when changing map style in a map with shape objects


hernanmateo:
Mapbox version: 6.0.0
React Native version: 0.49.3

I'm trying to switch between map styles in a Mapbox.MapView with a Mapbox.ShapeSource object with custom styles:

```js
render() {  
    var mapType = OUTDOOR_MAP_STYLE_URL;
    if (this.state.mapType == "satellite") {
      mapType = Mapbox.StyleURL.Satellite;
    } else if (this.state.mapType == "night") {
      mapType = Mapbox.StyleURL.Dark;
    }

    return (
      <View style={{backgroundColor: "white", flex: 1}}>
          <Mapbox.MapView
            ref={(c) => this._map = c}
            id="mainMapView"
            style={styles.map}
            maxZoomLevel={20}
            showUserLocation={true}
            styleURL={mapType}
            onDidFinishLoadingMap={this._onMapLoaded.bind(this)}
            >
              <Mapbox.ShapeSource id='userLocationShape' shape={point}>
                 <Mapbox.CircleLayer id='userLocationInnerCircleCnt' style={innerCircleStyle}/>
              </Mapbox.ShapeSource>
          </Mapbox.MapView>)
}
```
When I try to change the mapType, the application crashes in RCTMGLStyle.m line 1052. In the case that I remove the shape object styles (it does not matter if it is a line or a circle, etc ...) the map changes its style without crashing but the shapes disappears from the view.

nitaliano:
I have this working locally now, the fix will be in next weeks release

nitaliano:
@hernanmateo PR is out let me know how it goes - https://github.com/mapbox/react-native-mapbox-gl/pull/753

lucasbento:
@nitaliano: PR works fine, just approved it! 😄

nitaliano:
Has been merged into master closing out

lucasbento:
@nitaliano: do you know when you will release `6.0.1`?

nitaliano:
@lucasbento I'm going to do a release at the end of every week unless an urgent bugfix is needed, so I'll publish `6.0.1` this Friday.

lucasbento:
@nitaliano: alright, thanks!

PR #753, opened 2017-11-13
Adds support for switching styleURL with layers

nitaliano:
Adds support for https://github.com/mapbox/react-native-mapbox-gl/issues/736

* Users can now switch to any style url and their sources and layers will automatically re-add for them.


lucasbento (approved):
Just tested and works nicely, awesome work! 😄

PR #793, opened 2017-11-20
Touchable Vector and Shape Source

nitaliano:
* Adds support for an onPress on `VectorSource` and `ShapeSource` when one of their layers are pressed

### DIFF
--- a/javascript/components/ShapeSource.js
+++ b/javascript/components/ShapeSource.js
@@ -1,7 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { NativeModules, requireNativeComponent } from 'react-native';
-import { toJSONString, cloneReactChildrenWithProps, viewPropTypes } from '../utils';
+
+import {
+  toJSONString,
+  cloneReactChildrenWithProps,
+  viewPropTypes,
+  isFunction,
+} from '../utils';
+
 import resolveAssetSource from 'react-native/Libraries/Image/resolveAssetSource';
 
 const MapboxGL = NativeModules.MGLModule;
@@ -80,6 +87,12 @@ class ShapeSource extends React.Component {
      * you can specify an array of string names with assets as the key `{ assets: ['pin'] }`.
     */
     images: PropTypes.object,
+
+    /**
+     * Source press listener, gets called when a user presses one of the children layers only
+     * if that layer has a higher z-index than another source layers
+     */
+    onPress: PropTypes.func,
   };
 
   static defaultProps = {
@@ -126,6 +139,8 @@ class ShapeSource extends React.Component {
       id: this.props.id,
       url: this.props.url,
       shape: this._getShape(),
+      hasPressListener: isFunction(this.props.onPress),
+      onMapboxShapeSourcePress: this.props.onPress,
       cluster: this.props.cluster ? 1 : 0,
       clusterRadius: this.props.clusterRadius,
       clusterMaxZoomLevel: this.props.clusterMaxZoomLevel,
@@ -133,6 +148,7 @@ class ShapeSource extends React.Component {
       buffer: this.props.buffer,
       tolerance: this.props.tolerance,
       ...this._getImages(),
+      onPress: undefined,
     };
     return (
       <RCTMGLShapeSource {...props}>
@@ -143,7 +159,11 @@ class ShapeSource extends React.Component {
 }
 
 const RCTMGLShapeSource = requireNativeComponent(NATIVE_MODULE_NAME, ShapeSource, {
-  nativeOnly: { nativeImages: true },
+  nativeOnly: {
+    nativeImages: true,
+    hasPressListener: true,
+    onMapboxShapeSourcePress: true,
+  },
 });
 
 export default ShapeSource;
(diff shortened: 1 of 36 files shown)


javascript/components/ShapeSource.js
```diff
@@ -80,6 +87,12 @@ class ShapeSource extends React.Component {
      * you can specify an array of string names with assets as the key `{ assets: ['pin'] }`.
     */
     images: PropTypes.object,
+
+    /**
+     * Source press listener, gets called when a user presses one of the children layers only
+     * if that layer has a higher z-index than another source layers
```
lucasbento, line 93: How exactly do I specify a higher `z-index` on the `ShapeSource`?

nitaliano: It's based on your layers, let's say we have two sources that share the same feature collection of points.

```javascript
<ShapeSource id='test1' source={featureCollection} onPress={...}>
    <CircleLayer id='testLayer1' style={...} />
</ShapeSource>

<ShapeSource id='test2' source={featureCollection} onPress={...}>
   <SymbolLayer id='testLayer2' aboveLayerID='testLayer1' style={...} />
</ShapeSource>
```

the onPress for `test2` would be called, since it's layers has the highest z index order. 

Internally in Mapbox GL Native, there is an array called `layers` that is kept ordered by zIndex, so the 0 index is the backgroud and N-1 is the highest zIndex layer. When you call `queryRenderedFeaturesAtPoint` or `queryRenderedFeaturesInRect` that returns all of the layers in zIndex order N-1 to 0.

lucasbento (approved):
Works fine, thanks, @nitaliano!

lucasbento:
@nitaliano: I'm testing it and it calls only the last `ShapeSource`'s `onPress`, e.g. I have 4 polylines defined and pressing on any of them always calls the last defined one.

What I am doing is:

```jsx
handlePressPolyline = (id) => () => console.log('pressed', id);
```

```jsx
onPress={this.handlePressPolyline(id)}
```

lucasbento:
Also, sometimes the `onPress` doesn't get called, not sure if it's because of the `lineWidth` of the `LineLayer` or not, perhaps it would be good to increase the _size_ of the touchable element.

cc @nitaliano.

nitaliano:
@lucasbento I see what's happening I glanced at the code on the native side and I'm passing in all touchable sources when I look for the highest z-index layer, instead of passing in only the source that are at the point, so once I update to pass in the filtered touchable sources that should fix the issue.

For the hitbox, I can update the code to query by rect and by default I can make it 44x44 pixels, and what do you think about having a hitbox property on our sources to be able to make a custom hitbox?

lucasbento:
> @nitaliano: I see what's happening I glanced at the code on the native side and I'm passing in all touchable sources when I look for the highest z-index layer, instead of passing in only the source that are at the point, so once I update to pass in the filtered touchable sources that should fix the issue.

I see, that indeed should fix it.

> For the hitbox, I can update the code to query by rect and by default I can make it 44x44 pixels, and what do you think about having a hitbox property on our sources to be able to make a custom hitbox?

That seems to be a good idea, how would that look? props like `width` and `height` on `ShapeSource`?

nitaliano:
Yea, it could be something like `hitbox={{ x: 50, y: 50 }}` or `hitbox=[50, 50]`

lucasbento:
@nitaliano: first option looks better, why `x` and `y` though? shouldn't it specify `width` and `height`?

nitaliano:
yea, it can I'll make it `hitbox={{ width: number, height: number }}`

lucasbento:
That will look good!

lucasbento:
@nitaliano: do you have any idea of when you will be able to fix the `onPress`?

nitaliano:
@lucasbento I already have the code done for it I'll push the commits today

nitaliano:
@lucasbento changes have been pushed

nitaliano:
Published in 6.0.1 we can always add more updates to it if we run into issues

kaynz:
@nitaliano Thanks! That was fast :)

I'm getting this exception whenever I trigger the onPress of the ShapeSource: https://paste.xinu.at/lHKv9/

ShapeSource:
https://paste.xinu.at/ZiSG/xml

Symbols are rendered on the map & look fine, so I think my FeatureCollection is OK.

nitaliano:
I'm having trouble reproducing would you be able to post your geojson so I can run it through the Java code?

kaynz:
Sure: https://paste.xinu.at/7DE/json

I simply relinked via react-native unlink/link react-native-mapbox-gl to upgrade to 6.0.1. Is that enough?

nitaliano:
@kaynz try this out https://github.com/mapbox/react-native-mapbox-gl/pull/827

kaynz:
@nitaliano #827 fixes the error. Thanks :+1: 

Another question: How do I get the feature? I've got this object from onPress but not sure how to deal with it: https://paste.xinu.at/FTLJ/

kaynz:
@nitaliano any news on this?

nitaliano:
@kaynz look under `e.nativeEvent`

```
onSourceLayerPress (e) {
  const feature = e.nativeEvent.payload;
}
```

PR #797, opened 2017-11-22
Solves String.format issue while printing double values

alvaromb:
`String.format` was crashing under Android because `defaultValue` is a double and `%d` is for decimal numeric system, not decimal point.


nitaliano:
:+1: good catch thanks for the PR!

alvaromb:
Glad to have helped! 💣 💥  🐛

PR #806, opened 2017-11-27
More accurate headers

stabenfeldt:
As someone that hasn't worked a lot with Android previously, it's not obvious that the headers was referring to files. 

Eg:  **project:build.gradle** . Is that a function in a class or a relative path to a file?
I changed the headers to make that more clear.

### DIFF
--- a/android/install.md
+++ b/android/install.md
@@ -21,7 +21,7 @@ allprojects {
 }
 ```
 
-### app:build.gradle
+####  _$(PROJECT_ROOT)/android/app/build.gradle_
 
 Add project under `dependencies`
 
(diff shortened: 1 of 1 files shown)


android/install.md
```diff
@@ -2,7 +2,7 @@
 
 ## Gradle Setup
 
-### project:build.gradle
+#### $PROJECT_ROOT/android/app/build.gradle
```
nitaliano, line 5: I think you meant `$PROJECT_ROOT/android/build.gradle`?

stabenfeldt: I'm not sure. I found 2 `build.gradle` files:

```
$ find . -name build.gradle
./android/app/build.gradle
./android/build.gradle
```

Which one should it be? I'll update the patch accordingly. :-)

kaynz: `./android/app/build.gradle` is the right one.

stabenfeldt: Thanks, looks good! :)

nitaliano: The first path should point to `$PROJECT_ROOT/android/build.gradle` the one below this one is correct by pointing to `$PROJECT_ROOT/android/app/build.gradle`

PR #826, opened 2017-11-29
Image Source Support

nitaliano:
* Adds support for image overlays(no gif support yet, looking into it)

image_source


henrikra:
Nice!

PR #828, opened 2017-11-30
Update fitBounds to not use the Flight camera

nitaliano:
* Updates fitBounds on the MapView to not use the flight camera, to make it much faster
* fitBounds default duration has been changed to `0.0`


lucasbento:
reaction-Ls6ahtmYHU760

Issue #829, opened 2017-11-30
Add more information to offline packs


nitaliano:
We have more information on the native side, so we should add it to our JSON that we send up to the JS. 

From the tickets I've been reading it looks like we should add
* status(download status)
* countOfResourcesExpected
* maximumResourcesExpected
* countOfResourcesCompleted
* countOfBytesCompleted

josenaves:
Hey @nitaliano, for getPack, we have something like this:

```
{ pack: { 
    metadata: '{"name":"SFO_SATELLITE_0"}',
    bounds: [ [ -121.530762, 38.61687 ], [ -121.536255, 38.612578 ] ] 
  }, _metadata: null }
```

What about this:

```
{ pack: { 
    metadata: {
      "name":"SFO_SATELLITE_0",
      "downloadStatus": 1,
      "countOfResourcesExpected": 70000,
      "maximumResourcesExpected": 1232123432,
      "countOfResourcesCompleted": 3500,
       "countOfBytesCompleted": 12122
    }',
    bounds: [ [ -121.530762, 38.61687 ], [ -121.536255, 38.612578 ] ] 
  }
 }
```

nitaliano:
@josenaves looks great!

josenaves:
@nitaliano it turns out that we have a non exposed native method called getStatus.
So, I'm exposing it as getPackStatus method on JS side.

I had a hard time testing it. But, it's working

I'm opening a PR right now. I know that documentation and tests are missing for it.

Could you please take a look at my PR ?
https://github.com/mapbox/react-native-mapbox-gl/pull/836

nitaliano:
Closing this out it's in `6.0.2`

PR #836, opened 2017-12-04
[Android] Add getPackStatus function

josenaves:
This PR adds getPackStatus function for Android (Issue #829)

```MapboxGL.offlineManager.getPackStatus('packName')``` will return something like this:

```
status: {
  completedTileCount: 1313,
  requiredResourceCount: 2597,
  completedTileSize: 20150500,
  completedResourceSize: 93329811,
  completedResourceCount: 2597,
  downloadState: 0
}
```

### DIFF
--- a/javascript/modules/offline/offlineManager.js
+++ b/javascript/modules/offline/offlineManager.js
@@ -67,7 +67,7 @@ class OfflineManager {
    * const status = await MapboxGL.offlineManager.getStatus('packName');
    *
    * @param  {String}  name  Name of the offline pack.
-   * @return {Object} status Status object
+   * @return {Promise} Promise containing Status object
    */
   async getPackStatus (name) {
     if (!name) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by nitaliano** [comment]
First glance code looks good to me. I think we'll want to update the JS side unit test to reflect this new pack information. I can hop on the branch and add the iOS support and unit tests.

File: javascript/modules/offline/offlineManager.js [hunk 1]
```diff
@@ -60,6 +60,29 @@ class OfflineManager {
...
+  async getPackStatus (name) {
+    if (!name) {
+      return;
+    }
+
+    await this._initialize();
+    const offlinePack = this._offlinePacks[name];
+    
+    if (offlinePack) {
+      const status = await MapboxGLOfflineManager.getPackStatus(name);
```
nitaliano: We should wrap this in a try catch in case the native call rejects the promise. Maybe something like

```javascript
if (!name) {
  return null;
}

await this._initialize();

const offlinePack = this._offlinePack[name];
 
if (!offlinePack) {
   return null;
}

try {
   const status = await = MapboxGLOfflineManager.getPackStatus(name);
   return Promise.resolve(status);
} catch (e) {
   return Promise.reject(e); // so we don't lose the error
}
```

josenaves:
@nitaliano here is my code changes. Let me know if there is anything else more to do :)

nitaliano:
@josenaves great! I'll hop on the iOS stuff tomorrow and let you know when I push those changes and we can get this merged in. Thanks for jumping on this PR so quick btw!

josenaves:
URW ! Sounds great ! I'm looking forward to see this merged.

nitaliano:
I've been messing around with this and I think we can make some changes to the offline API separate of this. I'm currently working on top of this branch and I want to make it easier to listen status updates to packs that we are returning too, once I have those changes working I'll merge this in and have another PR up in the queue that has all the other changes

josenaves:
@nitaliano let me know if you need help on Android side.

PR #858, opened 2017-12-11
android texture mode option

hernanmateo:
Add TextureMode option to set instead of GLSurfaceView to avoid the android overlapping maps issue. The option is set in the RNMGLMapViewManager.java file because we have to set this options before the RNMGLMapView constructor.
 

Example usage:

```js
<Mapbox.MapView
        ref={(c) => this._map = c}
        [...]
        textureMode={true}
        >

### CODE REVIEW

**Comment by nitaliano** [comment]
Does this fix your problems with the navigators(I expect it would)? How is the performance with this?

**Comment by hernanmateo** [comment]
Yes this fix my problems, I have to test the performance in a device with a complex map. I'm just installing lets see..

File: android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java [hunk 1]
```diff
@@ -28,6 +30,8 @@
     public static final String LOG_TAG = RCTMGLMapViewManager.class.getSimpleName();
     public static final String REACT_CLASS = RCTMGLMapView.class.getSimpleName();
 
+    private boolean mTextureMode = false;
```
nitaliano: This variable would exist across all MapViews, since the manager is only instantiated on startup, all MapViews will share this code. For the most part this is fine, but I can see a use case where we might want to have a GLSurfaceView and TextureView exist in the same app.

I think we could approach this problem slightly different. We can make a completely native view manager for TextureView support so `RCTMGLTextureMapViewManager`, and from the JS side we can have a check in the render method like

```
render () {
   if (IS_ANDROID && this.props.textureMode) {
      return <AndroidTextureMapView ... />;
   }
   return <MapView ... />;
}
```

what do you think? I am a fan of exposing this variable.

hernanmateo: Perfect, I'm going to take a look during this week.

nitaliano: Great, I'll leave this open until then. If you need me to jump on a PR to handle this I would be happy to, this will help out a ton of people

hernanmateo:
Hi again @nitaliano,
let's see what do you think about this version, as you suggest I have separated MapView and AndroidTextureMap view, so I added 3 new files (AndroidTextureMapView.java, AndroidTextureMapViewManager.java and AndroidTextureMapView.js) they all inherit from the 'original' MapViews. 

Let me know if this was exactly what you were thinking.

nitaliano:
It's almost there. I don't think we shouldn't expose this in the public interface(`index.js`), `MapView.js` should make the decision to render this to component based on `IS_ANDROID && this.props.textureMode` and that should happen in `MapView.js` render function.

The native code looks good to me :+1:

hernanmateo:
Hi @nitaliano check the code and if its ok to you I hope we can merge.

nitaliano:
Great, LGTM I'll run this today to verify that it's working and we can merge it in

javascript/components/MapView.js
```diff
@@ -221,7 +238,9 @@ class MapView extends React.Component {
     logoEnabled: true,
     zoomLevel: 16,
     userTrackingMode: MapboxGL.UserTrackingModes.None,
+    //userLocationVerticalAlignment: MapboxGL.UserLocationVerticalAlignment.Center,
```
nitaliano: I'm assuming this is from your fork would you mind removing? and squashing down the commits

javascript/components/MapView.js
```diff
@@ -45,6 +47,11 @@ class MapView extends React.Component {
      */
     userTrackingMode: PropTypes.number,
 
+    /**
+     * The user location vertical alignment
+     */
+    userLocationVerticalAlignment: PropTypes.number,
```
nitaliano: Same would you mind removing

hernanmateo:
Ok done! In the last commit I added the comma as the original MapView.js... If it's necessary I can splash down again.

nitaliano:
Works great on Android there does seem to be a slight performance penalty on the 3d building example but overall seems good to me.

@hernanmateo one last thing I noticed is we get this warning on iOS `Warning: Native component for "RCTMGLAndroidTextureMapView" does not exist` we could probably wrap the `requireNativeComponent` for the map texture view in an if statement to prevent the warning

```
if (IS_ANDROID) {
const RCTMGLAndroidTextureMapView = 
  requireNativeComponent(ANDROID_TEXTURE_NATIVE_MODULE_NAME, MapView, {
    nativeOnly: { onMapChange: true, onAndroidCallback: true },
  });
}
```

hernanmateo:
Ok, I just pushed a version with this wrapped. I can't check now on iOS, but it's a little bit strange because I have been working all the morning on iOS and I haven't seen this warning. Any way, the wrapper makes sense, and I will take a look tomorrow to ensure that there wasn't any error with the last merge.

hernanmateo:
Ok, I just tested on iOS and I haven't received any warning. Let me know if you either.

nitaliano:
The warning only shows up in the Xcode logs, and it should fix it will merge this in today

nitaliano:
@hernanmateo I just tested it out with the if statement, it looks like it's creating the variable in a block scope and the variable is not being hoisted. We'll need to write it like this, it crashes otherwise

```
let RCTMGLAndroidTextureMapView;
if (IS_ANDROID) {
  RCTMGLAndroidTextureMapView = requireNativeComponent(ANDROID_TEXTURE_NATIVE_MODULE_NAME, MapView, {
    nativeOnly: { onMapChange: true, onAndroidCallback: true },
  });
}
```

hernanmateo:
Done! Isn't a big fix but a bug fix...

nitaliano:
Great thanks for the PR

hernanmateo:
Nice! Thanks to you

Issue #697, opened 2017-10-31
[v6] Add user initiated flag to regionDidChange callback


kristfal:
In order to keep redux state in sync with map camera state and without risking recursive loops, we'll need to know if it is a user initiated or programmatic event that triggered the region change.

Similar [GL-JS issue](https://github.com/mapbox/mapbox-gl-js/issues/1975), see [this blog post](https://blog.mapbox.com/mapbox-gl-js-in-a-reactive-application-e08eecf0221b) for a bit of backstory.

Addition: We should also be able to track if regionDidChange is triggered by userTrackingMode updates

nitaliano:
This is a great idea! I will look into getting this in the SDK

henrikra:
This is on very high priority for us too! 🌿

hannta:
Any news on this @nitaliano, I think this would be very useful feature for many use cases.

nitaliano:
I'm going to start looking into this next week. This is a very important feature IMO, I've wanted this feature in every app that I've built with this SDK so far. Once we finish adding support for being able to snapshot the map this is next the feature in line

nitaliano:
Does anyone have any opinions on what this will look like. I was thinking we just follow what GL JS has done and add an `originalEvent` property. Curious what others thought could go into the object at a minimum I was thinking

```
{
   type: 'zoom|rotate|pan|...' unsure what the event names will be
}
```

nitaliano:
I have a path forward on this I prototyped an idea to get this information into the region change events, expect a PR to land for this in the next few days.

henrikra:
Wow nice!

nitaliano:
WIP PR is out https://github.com/mapbox/react-native-mapbox-gl/pull/867 feedback would be greatly appreciated let's get this right. I'm eager to get this in so we can start focusing on location support

kristfal:
@nitaliano Had a look, and this is perf! With this, it's now possible to work around the location differences between Android and iOS as well. 

I do have one thought about `regionDidChange` and redux integration: Right now, using fitBounds with padding results in a regionDidChange that represents the derived viewport bounds and center, not the set bounds. This makes things tricky to track, say if we want to update the camera position by changing just the padding and reversing that change afterwards.

I'm wondering: Would it make sense if regionDidChange returned the original camera props for fitBounds, with the padding as a separate key rather than the derived center and bounds?

nitaliano:
Closing this out since we just merged the PR into master.

[...]

PR #867, opened 2017-12-12
Adds isUserInteraction to region change payload

nitaliano:
Adds support for https://github.com/mapbox/react-native-mapbox-gl/issues/697

* Region payloads now contain a boolean property called `isUserInteraction`
* [Android] Region update events are now throttled by 500 ms

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -6,6 +6,7 @@ import android.graphics.RectF;
 import android.location.Location;
 import android.os.Handler;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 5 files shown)


henrikra:
Nice!!! Does this fix also when screen is unmounted with MapView Android calls still `onRegionDidChange` event?

nitaliano:
@henrikra I don't think, what my plan is moving forward

* Next PR fix location on Android [userAlignment, userTracking]
* Next next PR update all map events to fire the same amount of times as the iOS SDK

and to use the iOS SDK as the source of truth for functionality. Once those two PR's are done I think that will leave this repo in a really good state.

henrikra:
Sounds great! Keep it up :D

kaynz:
Hi @nitaliano,
thanks for this PR. Just tested it - it's working perfectly :+1:

nitaliano:
Great, will work on getting this merged in

Issue #813, opened 2017-11-28
[v6] setCamera zoomLevel doesn't set zoom level


butchmarshall:
The zoom level remains fixed when calling setCamera (pitch, centerCoordinate, and duration all work fine).

```
map.setCamera({
	centerCoordinate: [44,76],
	zoomLevel: 16,
	pitch: 120,
	duration: 500,
});
```

nitaliano:
@butchmarshall the camera config calls it `zoom` instead of `zoomLevel`

butchmarshall:
Ah! Thanks!  Works like a charm!

Note: [docs](https://github.com/mapbox/react-native-mapbox-gl/blob/master/docs/MapView.md) should be updated.

PR #872, opened 2017-12-13
Fix: SetCamera config uses zoom instead of zoomLevel

jsaguiar:
#813


nitaliano:
Thanks for the PR will merge in today

PR #873, opened 2017-12-14
[ANDROID] blank info box displayed when annotation has no title

Tennen:
if an annotation has no title or snipped or callout, when it is selected, a blank info window is shown.
I think its related to this
screen shot 2017-12-14 at 4 18 29 pm

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -308,60 +308,54 @@ public class RCTMGLMapView extends MapView implements
 
     @Override
     public void onMapClick(@NonNull LatLng point) {
-        boolean isEventCaptured = false;
-
         if (mActiveMarkerID != -1) {
             for (String key : mPointAnnotations.keySet()) {
                 RCTMGLPointAnnotation annotation = mPointAnnotations.get(key);
 
                 if (mActiveMarkerID == annotation.getMapboxID()) {
-                    isEventCaptured = deselectAnnotation(annotation);
+                    deselectAnnotation(annotation);
                 }
             }
-        }
+        } else {
+            PointF screenPoint = mMap.getProjection().toScreenLocation(point);
+            List<RCTSource> touchableSources = getAllTouchableSources();
 
-        if (isEventCaptured) {
-            return;
-        }
+            Map<String, Feature> hits = new HashMap<>();
+            List<RCTSource> hitTouchableSources = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -405,11 +405,14 @@ public void selectAnnotation(RCTMGLPointAnnotation annotation) {
...
         if (id != mActiveMarkerID) {
             final MarkerView markerView = annotation.getMarker();
-            mMap.selectMarker(markerView);
+            String title = annotation.getTitle();
+            String snippet = annotation.getSnippet();
             annotation.onSelect(true);
             mActiveMarkerID = id;
-
             RCTMGLCallout calloutView = annotation.getCalloutView();
+            if (title != null || snippet !== null || (!markerView.isInfoWindowShown() && calloutView != null)) {
```
nitaliano: I think instead of this `annotation.getCalloutView` should just return `null` if it does not have a title.

What do you think?

Tennen: I'm not sure. will this effect the Callout component in an annotation?
such as, if the Annotation has no title or snippet, but it has a child Callout.

nitaliano:
@Tennen there's one difference I'm noticing between iOS and Android, once we fix that we can get this merged in. 

Once I have selected an annotation on Android then click on the Map(not on Annotation) it will fire the `MapView.onPress` and `Annotation.onDeselect` at the same time. iOS only calls `Annotation.onDeselect` and that is the desired behaviors of the SDK, mind updating?

Tennen:
@nitaliano the version I use in my project is v6.0.1, and the behavior is accurate..I'll try the latest version.

nitaliano:
@Tennen sorry forgot to mention that I remove the title, selected props and removed the callout to make it happen

Tennen:
@nitaliano I found this commit [dcf3c54](https://github.com/mapbox/react-native-mapbox-gl/commit/dcf3c54d35b9ca5692e134e5b2b36dc56d1b651e), it says if an annotation has no callout, deselect handler should not be called. But if the right behavior is calling `onDeselect` only, then this one should be reverted?
I updated my code, hope I'm not miss understanding...

nitaliano:
I'm going to close this out sorry we couldn't get it in `PointAnnotations` will be officially deprecated in our next major release and all focus will continue to be on Sources/Layers

Tennen:
@nitaliano it's ok, I'm using Sources/Layers now~

PR #897, opened 2017-12-21
Cross platform location

nitaliano:
* Polyfills `userTrackingMode` tracking on Android
* Adds `onUserTrackingMode` callback to MapView
* Enables userLocationVerticalAlignment support on iOS.
* Polyfills support for userLocationVerticalAlignment on Android
* Adds userLocationVerticalAlignment example
* [Android] Add support for updating map rotation based on user location for `followWithCourse` and `followWithHeading`

Bugfixes include
* Fixes Android not recentering camera when the developer sets a new tracking mode
* Fixes Android overriding `centerCoordinate` when `showUserLocation` is active
* Fixes Android not sending first region event
* Fixes iOS removing navigation icon anytime `followWithCourse` changes.
* [Android] Prevent user location layer from appearing under other layers

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -6,7 +6,7 @@ import android.graphics.RectF;
 import android.location.Location;
 import android.os.Handler;
 import android.support.annotation.NonNull;
-import android.util.Log;
+import android.util.DisplayMetrics;
 import android.view.View;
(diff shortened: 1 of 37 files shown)


nitaliano:
This PR is in pretty good shape, all that's left is to update the bearing of the map while in `followWithHeading` on Android. With that said if anyone has time to checkout the PR and lmk what you think that would be great!

nitaliano:
All features have been completed on this PR, there still might be subtle differences between the iOS and Android SDKs still but this is a HUGE improvement to what we currently have in master. I plan on merging this in this in tomorrow so this is your last chance to test it out before we get it in master.

Issue #965, opened 2018-01-18
[iOS] Cannot build project using installation via CocoaPods with `use_frameworks!`


JasonPan:
I'm installing the framework using CocoaPods as described in the iOS installation guide. The installation fails if I include `use_frameworks!` in the project's PodFile. The build errors that get shown are related to not being able to find React base files, such "React/RCTConvert.h file not found".

There are other dependencies in the project that are written in Swift, so leaving the option out of the PodFile isn't an option. I have tried

- changing my search paths
- deintegrating CocoaPods and reinstalling
- manually installing the framework by following the installation guide
- changing the import without the module name included - e.g. "RCTConvert.h"
- including the header files in the project's bridging header

Does anyone know what's going wrong? The project is using RN 0.51, Xcode 8 and Swift 3. When I leave out the project's Swift dependencies and `use_frameworks!`, the framework works completely (project builds, map gets shown).

nitaliano:
this repo hasn't been tested much with Swift codebases and cocoa pods mind creating a minimal repo and I can debug it from there?

JasonPan:
@nitaliano I've created a minimal project here: https://github.com/JasonPan/react-native-mapbox-issue

JasonPan:
@nitaliano As mentioned [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1525), looks like the React dependency just needed to be in the Podspec, I've made a PR that adds it in

nitaliano:
Fix has been merged into master

PR #984, opened 2018-01-22
[iOS] Add React dependency to Podspec

JasonPan:
Support older versions of Swift by allowing `use_frameworks!` in CocoaPods


JasonPan:
This resolves #965

kimhafr:
Tested this in our project as well, and it works! We had to add `use_frameworks!` ourselves after another dependency required us to use it

nitaliano:
Thanks for the PR!

Issue #999, opened 2018-01-27
(Android only) App freezes if you destroy activity and re-enter Application


cpetzel:
This issue is reproducible in the Example app in this repo. 

Steps to reproduce. 
1.) Run the example app
2.) Select the "Show Map" option and grant the location permission. 
3.) Observe that the app is requesting location in the status bar and displaying your location on the map. 
4.) Press back to dismiss the modal/map. (Observe that the application is still listening for location updates (status bar) even though the MapView is no longer visible)
5.) Press back again to destroy the activity. (Observe that the application is still listening for location updates (indicator in status bar)
6.) Relaunch the app (Observe a white screen. Pressing back again crashes the process)

I have verified that the LocationManager is notified to dispose itself in the react lifecycle callback [onHostDestroy()](https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java#L1014) Which deactivates the locationEngine. But somehow the Lost engine process is still alive and still asking for updates. 

After relaunching the app, This is the stack trace...
Seems like an IPC issue? The :lost process not communicating correctly? 

``` E/FusedLocationProviderServiceDelegate: Error occurred trying to report a new Location
                                                                         android.os.DeadObjectException
                                                                             at android.os.BinderProxy.transactNative(Native Method)
                                                                             at android.os.BinderProxy.transact(Binder.java:764)
                                                                             at com.mapzen.android.lost.internal.IFusedLocationProviderCallback$Stub$Proxy.onLocationChanged(IFusedLocationProviderCallback.java:129)
                                                                             at com.mapzen.android.lost.internal.FusedLocationProviderServiceDelegate.reportLocation(FusedLocationProviderServiceDelegate.java:95)
                                                                             at com.mapzen.android.lost.internal.FusionEngine.onLocationChanged(FusionEngine.java:191)
                                                                             at android.location.LocationManager$ListenerTransport._handleMessage(LocationManager.java:292)
                                                                             at android.location.LocationManager$ListenerTransport.-wrap0(Unknown Source:0)
                                                                             at android.location.LocationManager$ListenerTransport$2.handleMessage(LocationManager.java:244)
                                                                             at android.os.Handler.dispatchMessage(Handler.java:106)
                                                                             at android.os.Looper.loop(Looper.java:164)
                                                                             at android.app.ActivityThread.main(ActivityThread.java:6494)
[...]

Issue #925, opened 2018-01-05
App freezes, DeadObjectException


hannta:
On android im getting following app freezes, DeadObjectException, on some older Android devices.

I think this is similar issue, already reported on mapbox android sdk https://github.com/mapbox/mapbox-navigation-android/issues/522

At least in my case, this issue is reproducing only on react-native-mapbox-gl: 6.0.2, not 6.01.

```
01-04 16:49:48.866 7875-7875/com.my.app:lost E/FusedLocationProviderServiceDelegate: Error occurred trying to report a new Location
	 android.os.DeadObjectException
		 at android.os.BinderProxy.transactNative(Native Method)
		 at android.os.BinderProxy.transact(Binder.java:501)
		 at com.mapzen.android.lost.internal.IFusedLocationProviderCallback$Stub$Proxy.onLocationChanged(IFusedLocationProviderCallback.java:129)
		 at com.mapzen.android.lost.internal.FusedLocationProviderServiceDelegate.reportLocation(FusedLocationProviderServiceDelegate.java:95)
		 at com.mapzen.android.lost.internal.FusionEngine.onLocationChanged(FusionEngine.java:191)
		 at android.location.LocationManager$ListenerTransport._handleMessage(LocationManager.java:281)
		 at android.location.LocationManager$ListenerTransport.access$000(LocationManager.java:210)
		 at android.location.LocationManager$ListenerTransport$2.handleMessage(LocationManager.java:233)
		 at android.os.Handler.dispatchMessage(Handler.java:102)
		 at android.os.Looper.loop(Looper.java:135)
		 at android.app.ActivityThread.main(ActivityThread.java:5319)
		 at java.lang.reflect.Method.invoke(Native Method)
		 at java.lang.reflect.Method.invoke(Method.java:372)
		 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1016)
		 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:811)
```
 
react-native-mapbox-gl: 6.0.2
react-native: 0.51.0
device: android

vpavlushkov:
More details on the phone that experiences the issue:
 - device: Sony XPERIA M4 Aqua (E2303)
 - firmware: 26.1.A.3.111
 - Android version: 5.0

nitaliano:
[...]

PR #1001, opened 2018-01-27
fixes crash on Android re-enter app

cpetzel:
#999 #925 

The RCTMGLMapView has a LocationLayer that is registered with the locationEngine, and it never got removed when the view was torn down. This will tell the LocationLayer to stop when the view is detached. 

However, the Lost Engine is still active in the background. This patch simply lets you press back to close the app, and then re-enter without a crash

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -12,6 +12,7 @@ import android.util.DisplayMetrics;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.MotionEvent;
+import android.view.View.OnAttachStateChangeListener;
 
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactContext;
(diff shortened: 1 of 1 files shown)


cpetzel:
@nitaliano This seems to fix the crash in our app, but the LocationEngine is still active when the app is backgrounded (or the MainActivity is destroyed). 

With this change, you can successfully kill the activity and reenter. No more DeadObjectException.

nitaliano:
Awesome! I'll test this out. It might still be active in the background due to the telemetry service also using it.

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -169,6 +170,22 @@ public RCTMGLMapView(Context context, RCTMGLMapViewManager manager, MapboxMapOpt
         mHandler = new Handler();
 
         setLifecycleListeners();
+            
+        addOnAttachStateChangeListener(new OnAttachStateChangeListener() {
+            @Override
+            public void onViewAttachedToWindow(View v) {
+                if(mLocationLayer != null){
+                    mLocationLayer.onStart();
```
cpetzel, line 178: @nitaliano might want to wrap this onStart call with the same logic in [setReactShowUserLocation()](https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java#L700)

```setReactShowUserLocation(mShowUserLocation)```

henrikra:
We tried your branch and we still got the crash :/

cpetzel:
closing this in favor of #1010

PR #1010, opened 2018-01-30
Fixes for location engine (Android)

cpetzel:
#999 #925 

This PR is an improvement over #1001.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -9,6 +9,7 @@ import android.os.Handler;
 import android.support.annotation.NonNull;
 import android.text.LoginFilter;
 import android.util.DisplayMetrics;
+import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.MotionEvent;
@@ -169,6 +170,20 @@ public class RCTMGLMapView extends MapView implements
         mHandler = new Handler();
 
         setLifecycleListeners();
+    }
+
+    @Override
+    public void onWindowFocusChanged(boolean hasWindowFocus) {
+        super.onWindowFocusChanged(hasWindowFocus);
+        if (mLocationLayer == null) {
+            return;
+        }
+        if (hasWindowFocus) {
+            mLocationLayer.onStart();
+        } else {
+            mLocationLayer.onStop();
+        }
+
     }
 
     public void addFeature(View childView, int childPosition) {
@@ -991,8 +1006,8 @@ public class RCTMGLMapView extends MapView implements
     }
 
     private void setLifecycleListeners() {
-        ReactContext context = (ReactContext) mContext;
-        context.addLifecycleEventListener(new LifecycleEventListener() {
+        final ReactContext reactContext = (ReactContext) mContext;
+        reactContext.addLifecycleEventListener(new LifecycleEventListener() {
             @Override
             public void onHostResume() {
                 if (mShowUserLocation && !mLocationManger.isActive()) {
@@ -1013,6 +1028,7 @@ public class RCTMGLMapView extends MapView implements
             public void onHostDestroy() {
                 dispose();
                 onDestroy();
+                reactContext.removeLifecycleEventListener(this);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -171,6 +172,20 @@ public RCTMGLMapView(Context context, RCTMGLMapViewManager manager, MapboxMapOpt
         setLifecycleListeners();
     }
 
+    @Override
```
cpetzel, line 175: Instead of listening for window attachment changes, lets hook into the actual visibility of the window. This works for both backgrounding, and killing an activity (onStop vs onDestroy)

android/rctmgl/src/main/java/com/mapbox/rctmgl/location/LocationManager.java
```diff
@@ -110,6 +110,7 @@ public void disable() {
         if (locationEngine == null) {
             return;
         }
+        locationEngine.removeLocationUpdates();
```
cpetzel, line 113: as per the documentation, you need to call ```removeLocationUpdates()``` before you deactivate.

cpetzel:
@henrikra I think I may have fixed it in this branch. Want to give it a shot?

cpetzel:
Note: I had to add textureMode={true} to my map instances

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -1013,6 +1031,7 @@ public void onHostPause() {
             public void onHostDestroy() {
                 dispose();
                 onDestroy();
+                reactContext.removeLifecycleEventListener(this);
```
nitaliano, line 1034: should this line be moved to the `dispose` method?

android/rctmgl/src/main/java/com/mapbox/rctmgl/location/LocationManager.java
nitaliano: 👍 great catch

nitaliano:
I'm going to test this out more throughly before I merge this in. I had one minor comment, I think ideally we would want this to work without textureMode, but it's better than nothing for now

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
cpetzel: Because dispose is a public method, we wouldn't want anyone calling that and then our MapView never receives lifecycle callbacks anymore... 

I chose to put it in here because this mapView instance should only ever remove lifecycle events from react whenever the host is destroyed, not whenever someone else wants to call dispose()...

cpetzel:
I think these changes are needed regardless of textureMode. Perhaps with these changes, it could help uncover why textureMode is desired?

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
nitaliano: This is a good point, I agree that we should leave it out

nitaliano:
Thanks for the PR @cpetzel I'm going to merge this in and debug the issue further!

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
cpetzel: @nitaliano, After pondering about this further, I think that you are right. If React tells us to drop our mapView instance (```onDropViewInstance(mapView)```), and we don't remove the lifecycle listeners, we could see a nasty bug if the user background and resumes the app (hits the lifecycle callbacks).

The 'disposed' mapView instance would receive the onPause and onResume callbacks which would reenable (recreate and start) the location manager. We would not want this to happen. 

So I think we should also call ```reactContext.removeLifecycleEventListener(this);``` in the dispose method alongside calling it in ```onHostDestroy()```

Do you want to try and test this change? or should I open another PR for this?

PR #1011, opened 2018-01-30
Adds support for onUserLocationUpdate

cpetzel:
Instead of implementing our own location provider on both Android and iOS (the react geolocation watcher for Android is currently broken), we thought it would be better to have one source of truth for the user's location. 

This PR adds the ability to listen for location changes in javascript, coming from the LocationEngine callbacks in native code. 

A new prop is declared in the MapView - ```onUserLocationUpdate()``` which adheres to the [Web's GEO spec](https://dev.w3.org/geo/api/spec-source.html#position)

Example usage...

```
<MapboxGL.MapView
    onUserLocationUpdate={(location) => {
        console.log("received location", { location });
    }}
    showUserLocation={true}
/>
```

with the location object looking like this

```
{ location:
    { timestamp: 1517273596854,
    coords: { 
        speed: 0,
        heading: 0,
        accuracy: 19.354000091552734,
        altitude: -22.200000762939453,
        latitude: 37.7782095,
        longitude: -122.4087152 
        } 
    } 
}
```

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -145,6 +145,7 @@ public class RCTMGLMapView extends MapView implements
             if (mUserTrackingState == UserTrackingState.POSSIBLE || distToNextLocation > 0.0f) {
                 updateUserLocation(true);
             }
+            sendUserLocationUpdateEvent(nextLocation);
         }
     };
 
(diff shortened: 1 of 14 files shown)


cpetzel:
@nitaliano 
This is currently Android only, but we would finish the iOS feature if you think that this functionality can be added and maintained...

nitaliano:
@cpetzel I'm going to pull this branch down locally and add the iOS portion of it so when I do that I'll close this PR out and reopen the other one and tag you on it, thanks for the PR and this is something we will maintain!

cpetzel:
Awesome! Thank you.

cpetzel:
We would also like the ability to receive location updates even if showUserLocation is set to true. Could you do something like "if there is a function defined on the prop onUserLocationChange, then be sure to start the location engine and pass along updates" ?

seddonm1:
This is a really helpful feature. I thought I must have missed something from the docs when I couldn't find it.

Hope it can get into 6.0.3

EricPKerr:
@cpetzel thanks for this - excited to see it merged!  Is it possible to get the `onUpdateUserLocation` event without having `showUserLocation={true}` if we're showing our own UserLocation view?  The issue is that we want to get the location updates natively but snap the noisy-ish LatLon readings to a GeoJSON LineString (the user's route) if it's within a few meters, but the mapbox library isn't _aware_ of this.  Right now we're using the geolocation APIs, but would prefer to have a single cross-platform source of location updates and the fully native APIs appear more frequent and accurate.

cpetzel:
Yes, that was on our roadmap too. I will try and update this PR with that functionality for Android.

nitaliano:
@EricPKerr @cpetzel I think we should leave this PR as is for right now, I like the idea of breaking out the location from the MapView. If we wanted to do that I think it should be a module `MapboxGL.locationManager`, I'm also going to be pushing the iOS side of things to this PR shortly so let's say this is v1 and v2 will be the location manager

nitaliano:
I have added support for the iOS side, and made an example for it would someone mind looking over my code? then we can rebase with master and get this merged in

EricPKerr:
Works well on iOS - example is clear too.

nitaliano:
Great I think this is ready to merge in today, thanks again @cpetzel!

cpetzel:
awesome! I like the idea of a future MapboxGL locationManager. Thanks for getting this in.

Daavidaviid:
Hey, I can't get it to work on Android.. Are you sure it's ready ?

Issue #991, opened 2018-01-25
Cannot set MetaData on a new OfflinePack


atarsha:
In V5 it was possible to set additional metadata on an offline pack. This was very useful for various reasons.

Looking at https://github.com/mapbox/react-native-mapbox-gl/blob/bf1ce26495dff3234c9266fb8964b4ad2d9f1543/javascript/modules/offline/OfflineCreatePackOptions.js#L13 it looks like adding a parameter on the _makeMetadata() method and passing `options.metadata` to it from the constructor, you could add it to the property.

nitaliano:
Yea, this has come up a few times, and it will be added back. If you need to store data now I recommend using AsyncStorage and using the pack name as the key to store your data. I'll update this ticket once we add it back. The native code is already setup to handle the unstructured data the JS side just needs to be updated

nitaliano:
Closing this out as fix has been merged into master

PR #1017, opened 2018-02-01
Cannot set MetaData on a new OfflinePack

RobertSasak:
Fixing issue https://github.com/mapbox/react-native-mapbox-gl/issues/991

### DIFF
--- a/javascript/modules/offline/OfflineCreatePackOptions.js
+++ b/javascript/modules/offline/OfflineCreatePackOptions.js
@@ -35,8 +35,8 @@ class OfflineCreatePackOptions {
 
   _makeMetadata (metadata) {
     return JSON.stringify({
-      name: this.name,
       ...metadata,
+      name: this.name,
     });
   }
 }
--- a/javascript/modules/offline/OfflineCreatePackOptions.js
+++ b/javascript/modules/offline/OfflineCreatePackOptions.js
@@ -10,7 +10,7 @@ class OfflineCreatePackOptions {
     this.bounds = this._makeLatLngBounds(options.bounds);
     this.minZoom = options.minZoom;
     this.maxZoom = options.maxZoom;
-    this.metadata = this._makeMetadata();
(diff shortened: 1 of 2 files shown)


javascript/modules/offline/OfflineCreatePackOptions.js
```diff
@@ -33,8 +33,11 @@ class OfflineCreatePackOptions {
     return toJSONString(makeLatLngBounds(ne, sw));
   }
 
-  _makeMetadata () {
-    return JSON.stringify({ name: this.name });
+  _makeMetadata (metadata) {
```
atarsha, line 36: I think it would be a good idea to add some quick defensive code here with `delete metadata.name;` to prevent someone from over-writing the always-expected name value being normalized. Other than that, this looks great!

atarsha (changes requested):
I left a comment for a quick addition that would be good defensive code to prevent people shooting themselves in the foot.

javascript/modules/offline/OfflineCreatePackOptions.js
RobertSasak: You are right, I swapped *name* and *metadata* to prevent *name* from being over-written.

nitaliano:
This code does look good to me, before merging in I think we should add a quick unit test for this https://github.com/mapbox/react-native-mapbox-gl/blob/master/__tests__/modules/offline/OfflineCreatePackOptions.test.js#L4

you can test by running `npm test`

javascript/modules/offline/OfflineCreatePackOptions.js
atarsha: Yep that's a better way. So simple, why didn't I see it? :)

RobertSasak:
I extended existing test to also test metadata.

atarsha (approved):

nitaliano:
Just tested thanks for the PR @RobertSasak and thanks @atarsha for reviewing!

PR #1022, opened 2018-02-04
[Android/iOS]: Add support for mapView.getZoom()

### LINKED ISSUE #938: Get the current map zoom level.
I have a use case where I wanted to press a cluster and zoom the map by `(currentZoom + 2)` but I realised that it's not possible at the moment to retrieve the current `zoomLevel`.

mapbox-gl-js has this method for example:
https://www.mapbox.com/mapbox-gl-js/api/#map#getzoom

It would be great to have it, probably on the `ref` `this.map.getZoom()`

itsjgf:
getZoom

Closes #938.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -816,13 +816,10 @@ public class RCTMGLMapView extends MapView implements
 
     public void getZoom(String callbackID) {
         AndroidCallbackEvent event = new AndroidCallbackEvent(this, callbackID, EventKeys.MAP_ANDROID_CALLBACK);
-
         CameraPosition position = mMap.getCameraPosition();
 
         WritableMap payload = new WritableNativeMap();
-
         payload.putDouble("zoom", position.zoom);
-
         event.setPayload(payload);
 
         mManager.handleEvent(event);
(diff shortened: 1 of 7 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -814,6 +814,20 @@ public void queryRenderedFeaturesAtPoint(String callbackID, PointF point, Filter
         mManager.handleEvent(event);
     }
 
+    public void getZoom(String callbackID) {
```
nitaliano, line 817: mind removing the extra whitespaces in this method?

example/src/components/GetZoom.js
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      zoom: 9,
+    };
+
+
```
nitaliano: Same here we can get rid of some of the white spaces

example/src/components/GetZoom.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  onRegionDidChange = async () => {
```
nitaliano: We have normally been writing it like

```
async onRegionDidChange () => {}
``` 

mind updating so it matches the other places we use async/await

javascript/components/MapView.js
```diff
@@ -508,6 +508,21 @@ class MapView extends React.Component {
...
+   *
+   * @example
+   * const zoom = await this._map.getZoom();
+   *
+   * @return {Number}
+   */
+
+  async getZoom (){
+    const res = await this._runNativeCommand('getZoom');
+
```
nitaliano: Same here could you remove the white space

nitaliano (changes requested):
Other than a few nitpicks with code style. This PR looks great!

example/src/components/GetZoom.js
itsjgf: done 😛

nitaliano:
Thanks for updating will test this out then merge in

nitaliano (approved):

PR #1033, opened 2018-02-12
Fixes DeadObjectException

nitaliano:
* ~~Force TextureView on Android to prevent ANR/DeadObjectException since the Android SDK is expecting `onDestroy` to be called in the parents Activities `onDestroy` with a GLSurfaceView~~

~~One thing to note is there is a ticket open in `gl-native` to figure out a way to get the `GLSurfaceView` and `onDestroy` to work without needing to be called in the parent activity, so we are still keeping that code around because we expect it to work in future releases~~

See https://github.com/mapbox/react-native-mapbox-gl/pull/1033#issuecomment-366581754

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -97,6 +97,7 @@ public class RCTMGLMapView extends MapView implements
     private Handler mHandler;
     private LifecycleEventListener mLifeCycleListener;
     private boolean mPaused;
+    private boolean mDestroyed;
 
     private List<AbstractMapFeature> mFeatures;
     private List<AbstractMapFeature> mQueuedFeatures;
@@ -195,6 +196,12 @@ public class RCTMGLMapView extends MapView implements
     }
 
     @Override
+    public void onDestroy() {
+        super.onDestroy();
+        mDestroyed = true;
+    }
+
+    @Override
     public void onWindowFocusChanged(boolean hasWindowFocus) {
         super.onWindowFocusChanged(hasWindowFocus);
         if (mLocationLayer == null) {
@@ -272,6 +279,10 @@ public class RCTMGLMapView extends MapView implements
     }
 
     public synchronized void dispose() {
+        if (mDestroyed) {
+            return;
+        }
+
         ReactContext reactContext = (ReactContext) mContext;
         reactContext.removeLifecycleEventListener(mLifeCycleListener);
 
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -1,6 +1,8 @@
 package com.mapbox.rctmgl.components.mapview;
 
+import android.app.Activity;
 import android.content.Context;
+import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.PointF;
 import android.graphics.RectF;
@@ -14,6 +16,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.MotionEvent;
 
+import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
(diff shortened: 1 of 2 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -249,11 +250,23 @@ public AbstractMapFeature getFeatureAt(int i) {
...
+
         mLocationManger.dispose();
+
+        final RCTMGLMapView mapView = this;
+        reactContext.runOnUiQueueThread(new Runnable() {
+            @Override
+            public void run() {
+                mapView.onDestroy();
+            }
+        });
```
cpetzel: Because 'this' is a view, you could just post on its message queue. You can also just call the onDestroy method directly. 

```
post(new Runnable() {
            @Override
            public void run() {
                onDestroy();
            }
        });
```

nitaliano: o yeah duh, I was thinking I was in the manager here I'll call it directly 👍

cpetzel:
lgtm. However, I am curious why we have to always use a textureView on Android? this PR will now always call onDestroy on the view instance, so will that satisfy the GLSurfaceView issue regardless if it is in TextureMode or not? 

Could you elaborate a bit more?

nitaliano:
Currently the Maps Android SDK is expecting the `onDestroy` to be called in it's parents activity `onDestroy` with a GLSurfaceView, this is just the codepath that it takes inside the internals of the Maps SDK and then tries to cleanup the Map in the c++ portion of our codebase. That is where we are crashing, when using a TextureView the code takes a slightly different path preventing the crash.  Once the maps SDK fixes the ability to call the `onDestroy` outside of it's parent activity we will be able to update the JS side and allow the TextureView to be optional again. Once it's fixed upstream the only code changes that will need to be done is in the JS layer

cpetzel:
awesome. Thanks for that explanation.

henrikra:
I have update my test bug repo with your branch fix: https://github.com/henrikra/mapboxFreeezeBug

I recorded video with your fix and the repo from above: https://drive.google.com/file/d/1tuK5hqsYSPyl9MDjcMjnK_P0jqmh5ZL0/view?usp=sharing
The video was recorded on Huawei ALE-L21, Android version 6.0

Notes from Huawei ALE-L21 when testing in our main app:
- map was many times strecthed when coming back from a screen that has keyboard. Like this:
image
sometimes the strecthing went away when I moved the map and sometimes it stayed there
- you can see from the video that sometimes the modal on top was there since no touches where recognized at 0:10-0:17 and also at 0:33-0:40. So looks like sometimes the map just wont init it self or something


Notes from Nexus 5, Android version 6.0.1
- on first launch the map was totally black. on second launch map was there. So I guess the map did not update it self after it downloaded the map
- randomly maps are black
- if you put app to background and come back then the map is not black anymore
- on this phone when the map was streched it did not reset when you moved the map. Only if you put the app to background and came back then streching was gone

Happy news is that I could not see DeadObject crash. But looks like textureMode is giving us just other set of problems

nitaliano:
@henrikra I'll check it out. tobrun might have figured out a way to get the GLSurfaceView to work without having to call it from the Activity onDestroy and preventing the ANR from happening. I'm going to follow the steps he laid out and see what happens this PR might end up being adjusted with those new steps

nitaliano:
This is not going into `6.0.3` sadly, I'm going to be working with our Android team to see if we can really solve this issue without hacking around the problem, seeing some of the issues that are happening with the TextureView makes me believe it's not a good choice to force everyone to use it right now

henrikra:
This sounds good plan!

nitaliano:
I've managed to avoid the ANR from happening on an emulator and pushed those commits to this PR so others can also test...still need to do further testing on devices.

The last commit contains a custom map shadow node that will destroy the native map view when the shadow nodes dispose method is called.

vpavlushkov:
We have tried the commit `10b5c67`together with @henrikra on my phone Sony XPERIA M4 Aqua (Android 5.0) and it looks like we are back on square one now:
 - when returning to the map view, there is a black map in the view but the application freezes completely
 - after that, the only thing that works is to close the app from the apps manager, but that does not close the app completely. Next time I start the application after some delay I get Android notification that the application is not responding suggesting to close the app. Only after that, next start is successful. 
 - additionally, I saw before that when the app is in frozen state and a push notification comes to the phone, I still can open the corresponding app view through the deep link in the notification message.

It really looks like something gets disconnected/left behind when the user tries to return to the map view and until that resource is closed nothing happens... 🤔

nitaliano:
It's crashing when calling into the c++ codebase, when it's trying to destroy the native c++ mapview is when we are getting the ANR.

I have tested this on my Pixel so far, and it fixes the ANR that we were seeing in the exampleapplication. I don't have my Xperia on me right now but once I get back to my apartment later I will test it on there, it's a slightly higher version than the one being tested in here, but I'll see how it goes since I was able to reproduce those issues on it. Regardless if this solves 100% of the issues we are seeing it is a step in the right direction

nitaliano:
Just want to note I don't consider the DeadObject bug fixed. Now on master since the map will not stretch due to another bugfix textureMode is a good option now.

cpetzel:
👍

Issue #485, opened 2017-01-03
How can I change the language of the map?


shaharl:

ddolheguy:
Looks like unless I'm wrong this functionality is only available for Android (not iOS) and isn't available in this library.  You can raise a PR though :)

https://www.mapbox.com/help/change-language/

alexey-kuznetsov:
Is there any progress in this issue?

nitaliano:
This should be supported now that the runtime and data driven styling api is implemented.

henrikra:
Can this be reopened? Now there is api for the language: https://www.mapbox.com/help/change-language/#change-label-language-dynamically
So `localizesLabels` should be set to true.

nitaliano:
`mapView.propTypes.localizeLabels` have been merged into master going to close out

proffesor888:
Hey,
I'm setting localizeLabels to True but nothing changes. For example, default language on the phone is Russian. But town names and places are still in English. Is there any ways to change language of city names ?

PetrVasilev:
The same issue with @proffesor888

ancyrweb:
Same here, only on Android though.

totoblabla1:
Ребят, удалось найти решение?

thomasgauvin:
Has there been an update regarding this issue? localizeLabels does not work

benkeegan98:
has there been any resolution here? would like to be able to dynamically toggle localizeLabels, or at least toggle the language of the map by accessing the layer styles like is possible in Mapbox GL JS. is there any solution for this?

PR #1038, opened 2018-02-14
Turn locales on for iOS

henrikra:
Addresses #485 

This enables localisation on iOS and works like a dream 👗 . Documentation here: https://www.mapbox.com/help/change-language/#change-label-language-dynamically

I implemented only iOS because on Android localisation is way different. So if you guys want to help me how to do it then fine. But if not lets merge this? :)

### DIFF
--- a/android/rctmgl/build.gradle
+++ b/android/rctmgl/build.gradle
@@ -33,5 +33,6 @@ dependencies {
     }
 
     // Mapbox plugins
+    compile 'com.mapbox.mapboxsdk:mapbox-android-plugin-localization:0.1.0'
     compile 'com.mapbox.mapboxsdk:mapbox-android-plugin-locationlayer:0.3.0'
 }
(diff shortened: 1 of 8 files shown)


nitaliano:
I can handle the Android side, do you think this should be a boolean property on the map instead of just always enabling it?

henrikra:
That is one option to pass the boolean from JS yeah. Although I see localised case as the default. Like why someone doesn't want the labels to be same language as your system language :D?

nitaliano:
@henrikra I can't think of any reason why we someone wouldn't want this but maybe a reason exists, so when I add the Android side, I'll switch this to be a boolean just in case there is a use case to want this disabled.

I'll try to work on this PR this week since it's been hanging out for awhile

henrikra:
Good to hear

nitaliano:
@henrikra I made the changes want to look them over? I can rebase then merge this in after

henrikra:
Nice! I didn't know there was mapbox localize plugin for android! Looks good 👍

nitaliano:
great I'll get this merged in today

nitaliano:
thanks again @henrikra for getting this started!

PR #1040, opened 2018-02-14
[Android] Fix filters not being applied to a layer after switching styleURL

frankrowe:
On Android, when switching styleURL of the map, layers are removed and added back to the map. Filters are not correctly applied when this happens.

To test, add a ShapeSource with two layers, both with filters. Add a button to switch styleURL. 

I believe this is because RCTLayer stores a FilterList property. When this is parsed, the items are removed from the array. The next time the layer is added to the map, this FilterList is empty.

My suggestion to fix this is to store the Filter.Statement on the Layer, instead of the FilterList. This solved the issue for me in limited testing.


nitaliano:
@frankrowe I agree with the changes you made here glancing thru the code. I'll test this out after I release `6.0.3` tomorrow and work on getting this merged in

nitaliano:
Reviewed code more throughly and it looks great! I just tested it out on Android and everything is working great, going to test this out on iOS to make sure that things are behaving the same

Update:
iOS is behaving correctly here, thanks again for the PR!

Issue #920, opened 2018-01-01
Random crashes with custom png SymbolLayer marker


mtt87:
Currently testing only on iOS I see many random crashes in the app when there is a Map component using `ShapeSource` and `SymbolLayer` with a custom `.png`
```js
const markerBlack = require('../../images/pin_nero.png');

const layerStyles = MapboxGL.StyleSheet.create({
  icon: {
    iconImage: markerBlack,
    iconAllowOverlap: true,
    iconIgnorePlacement: true,
    iconSize: Platform.OS === 'android' ? 1 : 0.5,
  },
});

<MapboxGL.ShapeSource
  id="restaurant"
  shape={{
    type: 'FeatureCollection',
    features: [
      {
        type: 'Feature',
        geometry: {
          type: 'Point',
          coordinates: [lng, lat],
        },
        id: 1,
      },
    ],
  }}
>
  <MapboxGL.SymbolLayer id="singlePoint" style={layerStyles.icon} />
</MapboxGL.ShapeSource>;

```
It's definitely in the native code because the app is totally crashing (no RN red screen with the stack trace, no chrome debug stack trace), I'm not very proficient with native code and Xcode, this is what I can see
`Mapbox`___lldb_unnamed_symbol8632$$Mapbox:`
<img width="1128" alt="screen shot 2018-01-01 at 23 20 14" src="https://user-images.githubusercontent.com/2150927/34471457-aa0a57bc-ef4a-11e7-81e1-d625b9b10e0b.png">

mtt87:
With https://github.com/mapbox/react-native-mapbox-gl/issues/918#issuecomment-354847711

Seems like it's not crashing anymore 👍 

I'll do a bit more testing tomorrow and if there are no more crashes I'll close this issue 😄  thanks!

nitaliano:
@mtt87 just made a PR for it https://github.com/mapbox/react-native-mapbox-gl/pull/922

jsaguiar:
It seams to be fixed! thanks @mtt87

EricPKerr:
@nitaliano I am getting this error from the latest merged code in master.  It includes the merged https://github.com/mapbox/react-native-mapbox-gl/pull/922 PR:

```
Line 195:
[RCTMGLUtils fetchImage:_bridge url:styleValue.payload[@"value"] callback:^(NSError *error, UIImage *image) {
  if (image != nil) {
    dispatch_async(dispatch_get_main_queue(), ^{
[...]

PR #1080, opened 2018-03-01
Bugfix/ios remote image crashes

nitaliano:
Fixes https://github.com/mapbox/react-native-mapbox-gl/issues/920#issuecomment-368691849

* [iOS] Fixes random crash with remote images in symbol layer
* [iOS] Fixes crash when switching between styleURLs with a SymbolLayer rendered on the map with `iconImage` set.

Thanks @atarsha for the support and sample project!


atarsha (approved):
I went ahead and pulled these changes, since they aren't EXACTLY the same as the previous gists we used. Despite my best efforts, I did not reproduce a single crash. This is fantastic! Bravo!

PR #1100, opened 2018-03-07
Second "onCameraIdle" event rejected on Android

Gp2mv3:
There is an issue in the Android SDK, the "onCameraIdle" event is thrown twice. In the wrapper, there is a minimal delay set to avoid calling onRegionDidChange twice for the same event.
This is a workaround to avoid the second (and real) event to be dismissed because it's happening too fast.

In fact, the issue is really in the Android SDK but it seems to be corrected in a new version of the SDK. So this PR is just a Workaround before upgrading the Android SDK.


nitaliano (changes requested):
Code wise this looks good, one small nit could you update the indentations to match what's already in the code and I can test this out and get it merged in?

nitaliano (approved):

nitaliano:
@Gp2mv3 thanks for updating I'll test this out today

PR #1105, opened 2018-03-09
Fix a typo which causes onWillStartLoadingMap not get called.

studyro:
We found that the `onWillStartLoadingMap` in `MapView` not get called.

It's caused by a typo in `_onChange` in `MapView.js`. This PR fixes it.


nitaliano:
Thanks for the PR

PR #1117, opened 2018-03-15
Added option to force Orientation display on map

Gp2mv3:
This PR allows to display the Heading without having to follow the user position.


nitaliano:
@Gp2mv3 thanks for the PR, I agree that we would want to have the icons not be based on user tracking mode.

I invision a future where user location is not coupled with the MapView props and we'll have a component called `MapboxGL.UserLocation` where we would be able to listen for events, decide what stock icon we want, be able to add a custom icon. What do you think about that?

Gp2mv3:
Seems good ! I was also adding a function to center the map programmatically to the user position.

But it's indeed a good idea to decouple everything. Do you have a plan ?

nitaliano:
Sorry for the delay @Gp2mv3 this is the next PR to be merged in, I'm going to test this out later today the code looks good to me.

As for the plan this is what I think needs to happen.
1. Create `MapboxGL.locationManger` that would allow users to request user location without the need to have `showUserLocation`. This will also allow us to give the developer the option of using Google or Lost as their location engine on Android.

2. Create `MapboxGL.UserLocation` that is backed by a shape source and layers(so we can customize it), add support for user tracking modes, automatically hook into when the `MapboxGL.locationManger` is requesting updates and move icon based on tracking mode. (This is a big task)

3. Remove all user location props from `MapboxGL.MapView`

nitaliano (changes requested):
@Gp2mv3 would you mind removing the `centerMap` method from the Android portion of this PR? Would like to get in just the orientation icon changes and also could we clean up the commit history? It's always best to rebase instead of create a merge commit to keep the history clean and easier to understand, since we'll be creating a merge commit when it gets merged into master, if you need help I can help out with the git portion.

Gp2mv3:
I reverted the last commit, you should be able to merge.

nitaliano (approved):

nitaliano:
Thanks @Gp2mv3 👍

samdelacruz:
Is this likely to make it into the next release?

kristfal:
Closing this one as its stale.

PR #1119, opened 2018-03-15
Add support for resolving style values

nitaliano:
* Resolve style values so user's can use for instance `round` instead of `MapboxGL.LineCap.Round`
* Update style docs to be much more verbose


atarsha (approved):
Having the list of possible values in each of the various documentation updates will be so extremely beneficial!

I feel like the markup format leaves a little to be desired, but maybe in a future update. This is a dramatic improvement!

nitaliano:
We can polish up the markdown before this gets merged in. Do you know any examples of what would look good?

nitaliano:
Markdown was updated very slightly to make reading it easier, and a list of style props have been added above the style section so you can click and it will scroll to your style. Going to merge this in today

PR #1123, opened 2018-03-19
[Experimental] React native link support

nitaliano:
* Adds support for `react-native link @mapbox/react-native-mapbox-gl`

This is still experimental, would like to merge this into master with a note about this being experimental in the README.


henrikra:
I have very bad experiences using `react-native link` when doing large project. It basically never works.

nitaliano:
I agree that it is normally a diaster, but I have had success with it once I install a new dependency and only call react-native link packageName if rnpm is setup correctly which isn't always the case.

PR #1132, opened 2018-03-22
Fixes onRegionWillChange not firing on Android

nitaliano:
* Fixes `onRegionWillChange` on Android

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -146,6 +146,8 @@ public class RCTMGLMapView extends MapView implements
     private ReadableArray mInsets;
     private Point mCenterCoordinate;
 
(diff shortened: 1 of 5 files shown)


terryttsai:
@nitaliano I'm testing out master with this bugfix merged in on Android, and I'm still getting unexpected behavior, with `regionDidChangeDebounceTime={0}` and `regionWillChangeDebounceTime={0}`

When I pan the map without flicking the map, `onRegionDidChange` fires when I let go, and `onRegionWillChange` sometimes fires and sometimes doesn't. `onRegionIsChanging` doesn't get fired.

When I flick the map, `onRegionWillChange` gets fired, followed by a few calls of `onRegionIsChanging`, followed by `onRegionDidChange`, followed by a series of `onRegionIsChanging` calls while the map is panning by inertia, then ending with `onRegionDidChange`. 

What I don't expect is in the first case, `onRegionIsChanging` doesn't get called, and `onRegionWillChange` doesn't get called 80% of the time. And in the second case, `onRegionDidChange` gets called twice. 

This behavior is consistent among the two android phones I tested.

Issue #1015, opened 2018-01-31
Update Mapbox.framework fetch script to work with yarn on updates


alexgvozden:
Hi, 

I am using MapBox in React Native app
At first it compiles properly and runs - everything works 
but after some time on rerun it drops a load of errors
then you have to remove node_modules all the time and reinstall it in order for app to compile

<img width="1193" alt="screen shot 2018-01-31 at 10 11 58 am" src="https://user-images.githubusercontent.com/127109/35614644-1a17b3ba-0670-11e8-97c7-6981a5c2e515.png">

nitaliano:
What version of the mapbox framework is under your node_module? I would try a fresh install those symbols should exist

alexgvozden:
Hi

release is 6.0.3rc 

I do fresh install every day because of this error, after compiling this error comes haunting.
On removing node_modules and installing again it works so it's something in compilation or similar that removes the files or adds additional stuff

Thanks

nitaliano:
are you using yarn or npm and what version are you using? I haven't run into this with the example app yet

lesserj:
I've had similar and consistent issues with this library where the build breaks after some time with errors similar to above.  I have to pretty regularly run the following:
yarn remove @mapbox/react-native-mapgox-gl
yarn add @mapbox/react-native-mapbox-gl

I think it works until the xcode cache is removed.  Closing and reopening xcode seems to cause the problem as does the occasional yarn upgrade.  I'm currently on 6.0.3-rc1, but have had similar problems with past versions as well.

nitaliano:
I've been using `npm` so I'll checkout `yarn` and see what's happening

jmcneese:
I have this issue too, using yarn.

nitaliano:
This will be fixed in version `6.0.4`

alexgvozden:
awesome @nitaliano  
my mornings are going to be more beautiful now :) 

almost done with the app I'll share it once out

jaltin:
Hi @nitaliano,

I am using 6.1.0 version and I just ran into this issue. Was it not fixed in the 6.0.4 version (as per your comment above) or has it perhaps reappeared?

@lesserj's solution/workaround works for me for now, but would be great to not have to do it every time the problem reappears.

Thanks!

nitaliano:
[...]

PR #1151, opened 2018-04-02
Ability to use PlistBuddy to determine framework version

fwal:
For me I had a case where the version in `.framework_version` and the actual version of the framework was different, causing the script to think that it already had the latest version.

Here it looks at the version inside the framework itself by using `PlistBuddy` if available.

### DIFF
--- a/scripts/download-mapbox-gl-native-ios.sh
+++ b/scripts/download-mapbox-gl-native-ios.sh
@@ -3,7 +3,11 @@
 cd ios/
 
 VERSION=$1
-CURRENT_VERSION=$(cat .framework_version)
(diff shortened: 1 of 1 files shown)


nitaliano:
Thanks for the PR! I'll test this out later today

nitaliano:
@fwal I think this will also fix this issue that we were going to look into for `6.1.1` https://github.com/mapbox/react-native-mapbox-gl/issues/1015

fwal:
@nitaliano That's great!

nitaliano (approved):
Tested this out and it works great, thanks again!

Issue #1014, opened 2018-01-30
Add fromScreenLocation method


kristfal:
Hey,

a RN bridge to the `fromScreenLocation` / `toCoordinateFromView` methods would be very nice to have. The method should accept the RN PanResponder x0, y0 method and return the underlying lat, lon coordinate.

Our current workaround is to get screen bounds and calculate coordinates based on screen fractions from the bounds. In some ways, this approach may be more performant than using the native methods as it won't require an additional pass over the bridge.

nitaliano:
We have this method right now https://github.com/mapbox/react-native-mapbox-gl/blob/master/docs/MapView.md#getpointinviewcoordinate

hampelm:
Was just looking at this myself. According to the docs, the current `getPointInView` method takes a lat lng and returns the point in view. It sounds like what @kristfal is looking for is the inverse, a method that takes map coordinates and turns them to lat-lng. When I pass `getPointInView` the `screenPointX, screenPointY` from `onPress`, it returns [0,0]. 

Additionally, when I use `getPointInView` as shown in the docs I get this error (but the async function still evaluates). Not sure yet if it's my code (I do have the `this._map` ref set and working) or something else. Most likely the former; if I can isolate it I'll open an issue but wanted to mention it. 

```
RCTJSONStringify() encountered the following error: Invalid number value (NaN) in JSON write
reactConsoleErrorHandler @ ExceptionsManager.js:73
console.(anonymous function) @ console.js:32
console.error @ YellowBox.js:71
logToConsole @ RCTLog.js:48
logIfNoNativeHook @ RCTLog.js:31
__callFunction @ MessageQueue.js:353
(anonymous) @ MessageQueue.js:118
__guardSafe @ MessageQueue.js:316
callFunctionReturnFlushedQueue @ MessageQueue.js:117
t @ RNDebuggerWorker.js:1
```

kristfal:
Hey! Sorry for the unclear issue description. Yes, what we currently lack is a method for calculating lat, lon from screen coordinates.

terryttsai:
I second this. Currently, I can use [getCenter](https://github.com/mapbox/react-native-mapbox-gl/blob/master/docs/MapView.md#getcenter) to get the coordinate at the center of the map view, but sometimes I want to get the coordinate at a different part of the map view, like with some Y offset. So it would be nice to have a `getCoordinateFromPoint` method.

[...]

PR #1154, opened 2018-04-03
Add method to get geographic coordinate from view point

frankrowe:
Adds the `getCoordinateFromView(point)` method to the mapview, which returns a coordinate from a point in the map view.

Addresses https://github.com/mapbox/react-native-mapbox-gl/issues/1014


nitaliano:
Thanks for the PR @frankrowe I'll test this out tomorrow and we can get this merged in this week!

Issue #1160, opened 2018-04-06
Resetting filters (at runtime) back to null does not have any affect


stdavis:
I have an app that allows the user to interactively filter the operational data. Wiring up the filter state to `CircleLayer.filter` works great until I try to reset the filter back to `null`. When I try to do this, the layer does not update and continues to be filtered by the previous values.

To reproduce:
- Display a CircleLayer with the initial filter set to `null`.
- Add a filter (e.g. `["any",["==","Type","p"]]`)
- Observe that only `p` type features are displayed on the map.
- Set the filter back to `null`.
- Observe that only `p` type features are still displayed on the map. It should show all features.

I believe that the issue is here: https://github.com/mapbox/react-native-mapbox-gl/blob/a8d68af7efd6e55869ae3ec4fe0670ec1a58cc3d/ios/RCTMGL/RCTMGLLayer.m#L82-L84

If I remove the if block then the layer behaves as expected.

I also believe that this affects all layer types, not just `CircleLayer`.

Android appears to behave the same way although I haven't tracked down the native code yet.

Am I using the `filter` property incorrectly or is this a bug? If it's a bug, I'd be willing to take a crack at a PR if you can point me in the right direction.

Thanks for the amazing project.

```
"@mapbox/react-native-mapbox-gl": "^6.1.0",
"react": "^16.3.0",
"react-native": "0.54.2",
```

nitaliano:
@stdavis yes great point, all of the layers are doing a null check here and we should absolutely remove this check.

On Android it's little more verbose

There is a base layer class
[RCTLayer](https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/layers/RCTLayer.java#L237) this class has some base methods `updateFilter`, `buildFilter`, and `setFilter`.

Every layer class overrides the `updateFilter` method to call into `setFilter`(this method is on the actual native Android mapbox layer) you can see this in [RCTMGLFillLayer](https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/layers/RCTMGLFillLayer.java#L25)

I think we need to remove this null check https://github.com/mapbox/react-native-mapbox-gl/blob/master/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/styles/layers/RCTLayer.java#L164

[...]

PR #1162, opened 2018-04-06
remove check for nil filter

stdavis:
Related to #1160 

This fixes the problem in iOS for me. However, Android is throwing this error:
image

I'm not experienced with Java but did a bit of digging. Couldn't find out where the issue was. I almost think that it's upstream in the mapbox sdk or something.


nitaliano:
@stdavis thanks for the PR! I'll test this out and we can work on getting this merged in

kristfal:
@stdavis hey! I'd be happy to merge if you can add the android functionality for this so iOS and Android behave the same. 

In the interim, I'll close the PR, but feel free to reopen.

PR #1164, opened 2018-04-08
Fix typo in StyleSheet.md

mohammad7t:
Copying and pasting code snippet with this typo causes all colors to be rendered black without any error. It was too hard to figure out the cause was this typo.


nitaliano:
Thanks for the PR

Issue #807, opened 2017-11-27
iOS Memory Leak


stephenheron:
Hi,

I am having a memory leak issue that I am having trouble tracking down.

I am using the MapBox map a fair amount and when you navigate deeply down through the app and then back up the memory usage continues to rise. 

I have a video showing the issue here: https://youtu.be/uy1kGgzM50Q 

As you can see when you drill down into the app the memory rises. This is to be expected but when you go back up to the top level the memory usage does not drop down to reasonable levels. After doing this only 3 times the memory usage is over 1GB. The same behaviour happens on the phsyical device as well.

If I disable the map component then I don't see the above behaviour.

I am using react navigation as the navigation library. I know this could be quite difficult to solve so any more information or anything I should try then just let me know.

Thanks,
Stephen

nitaliano:
Is componentWillUnmount called at all when using react navigation? I've known from past use using react navigation it might not be unmounting the components correctly. I've profiled the example application before with the modal and it unmounts the Map and memory goes down

stephenheron:
Hi,

componentWillUnmount does seem to be called correctly. 

However I might have got to the bottom of the issue. I took the simplest map example and dropped it in place of my map and the memory leak went away. I then added functionality back in bit by bit until the memory leak appeared and it seems to be something todo with the geoJSON. 

Sometimes my map has geoJSON and other times it does not so I had some code to handle that.

This is what I had before:
```
this.state = {
   geoJSON: {},
};

renderGeoJSON() {
      return (
      <MapboxGL.ShapeSource id='routeLineSource' shape={this.state.geoJSON}>
        <MapboxGL.LineLayer id='routeLineLayer' style={layerStyles.routeLine} />
      </MapboxGL.ShapeSource>
    )
}
```
However when I updated the geoJSON to be null and then check it for null it seems to make the memory leak disappear.

```
this.state = {
   geoJSON: null,
};

renderGeoJSON() {
    if(this.state.geoJSON == null) {
      return null;
    }

    return (
      <MapboxGL.ShapeSource id='routeLineSource' shape={this.state.geoJSON}>
        <MapboxGL.LineLayer id='routeLineLayer' style={layerStyles.routeLine} />
      </MapboxGL.ShapeSource>
    )
  }
```
Seems a tad strange to be honest. I have not had enough time to test it properly on a real device but on first glance it does seem to make the issue disappear. 

Will test a bit further and see if it has truly fixed the issue.

Thanks,
[...]

PR #1165, opened 2018-04-09
Fixed iOS memory leak discussed in #807

studyro:
There're two issues in the original implementation:

1. In `- [RCTMGLMapView invalidate]`, it mutates the Array `_reactSubviews` during iteration, so not every element in `_reactSubviews` can be handled by `[RCTMGMapView removeReactSubview:]`.<br/> And the leaked `RCTMGLSource` instances strongly reference `RCTMGLMapView`, which causes the retain cycle never get released.
2. By calling `[self removeReactSubviews:(UIView *)_reactSubviews[i]];` in `- [RCTMGLMapView invalidate]`, `_reactSubviews` get changed. But [React Native depends on this](https://github.com/facebook/react-native/blob/bfcfe7961db0970e2575eafe2f3c9c668bd8940d/React/Modules/RCTUIManager.m#L436) to traversely clean `_viewRegistry` in `RCTUIManager`.

We just need to clean internal state in `- [RCTMGLMapView invalidate];`, leaving the `_reactSubviews` to be handled by `RCTUIManager`.

### DIFF
--- a/ios/RCTMGL/RCTMGLMapView.m
+++ b/ios/RCTMGL/RCTMGLMapView.m
@@ -37,7 +37,7 @@ static double const M2PI = M_PI * 2;
         return;
     }
     for (int i = 0; i < _reactSubviews.count; i++) {
-        [self removeReactSubview:(UIView *)_reactSubviews[i]];
+        [self removeFromMap:_reactSubviews[i]];
     }
 }
 


nitaliano:
Awesome, thanks for the PR! I'll test this out and we can get this merged in.

nitaliano:
This works great thanks again!

Issue #1106, opened 2018-03-12
IllegalArgumentException crash on Android


hannta:
Im seeing several following crashes on Google Play console.

```
java.lang.IllegalArgumentException: 
  at com.mapbox.mapboxsdk.maps.renderer.MapRenderer.onSurfaceChanged (MapRenderer.java:67)
  at com.mapbox.mapboxsdk.maps.renderer.textureview.TextureViewMapRenderer.onSurfaceChanged (TextureViewMapRenderer.java:48)
  at com.mapbox.mapboxsdk.maps.renderer.textureview.TextureViewRenderThread.run (TextureViewRenderThread.java:269)
```
react-native-mapbox-gl: 6.1.0
react-native: 0.52.2
device: android

henrikra:
This is our number one crash at the moment :(

hannta:
@nitaliano do you have any ideas about this?

henrikra:
@nitaliano we are having tons of this crash 💣 Do you have an idea for this? 💡

nitaliano:
Is there anymore information on this or do you know how to reproduce this?

tobrun:
This can only occur as part of the `onSurfaceChanged` callback. We haven't had reports of this issue upstream in gl-native. @hannta / @henrikra , do you see any additional error message to this `IllegalArgumentException`? eg. `fbWidth cannot be negative.` or `fbWidth cannot be greater than 65535.` I will create an issue upstream and link it back here, any additional information would be highly appreciated!

henrikra:
Yeah the logs say `fbWidth cannot be negative`

```
java.lang.IllegalArgumentException fbWidth cannot be negative. 
    MapRenderer.java:67 com.mapbox.mapboxsdk.maps.renderer.MapRenderer.onSurfaceChanged
    TextureViewMapRenderer.java:48 com.mapbox.mapboxsdk.maps.renderer.textureview.TextureViewMapRenderer.onSurfaceChanged
    TextureViewRenderThread.java:269 com.mapbox.mapboxsdk.maps.renderer.textureview.TextureViewRenderThread.run
```

Does this help you?

nitaliano:
@henrikra yes it does, just for reference we opened an issue in gl-native as well https://github.com/mapbox/mapbox-gl-native/issues/11549

Rnorback:
Any updates on this?

nitaliano:
@kmagiera can you make a PR with that fix?

[...]

PR #1171, opened 2018-04-16
[Android] Prevent onLayout from being called while app is in paused state

nitaliano:
Fixes https://github.com/mapbox/react-native-mapbox-gl/issues/1106

This will prevent us from trying to update the MapView in a paused state with a negative width and height

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -527,6 +527,13 @@ public class RCTMGLMapView extends MapView implements
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


henrikra:
Problem for me is that I can't reproduce the bug but I see lot of those crashes on production. So I am willing to put this to production and see if the crash count goes down :)

nitaliano:
Ok, I'll just merge this in and I'll publish a beta on npm tomorrow so you can test it out

henrikra:
Alright! Right when you have made new release I can put it to production :)

nitaliano:
@henrikra I just published `6.1.2-beta`

nlindroos:
@henrikra are you still receiving reports of this crash with `6.1.2-beta`? The crashes seem to occur for many of our users in production.

PR #1188, opened 2018-05-08
[fix] fix pointannotation flying to left top when it's layout changes

Tennen:
if children of annotation have unknown size, when the content inside changes, reactSetFrame is called, then it will fly to the left top.  

I have an example.
image   
by the way, when will annotation view be implemented?  : )


nitaliano:
Thanks for the PR!

PR #1198, opened 2018-05-14
[fix] Fix isUserInteraction for onRegionWillChange

### LINKED ISSUE #980: onRegionWillChange isUserInteraction always false
This is potentially related to https://github.com/mapbox/react-native-mapbox-gl/issues/901 but it appears that the `isUserInteraction` on the `onRegionWillChange` callback is always false (when user drags/pinches/etc).  This was tested and verified with the **6.0.3-rc1** release.

EricPKerr:
This fixes https://github.com/mapbox/react-native-mapbox-gl/issues/980

`regionWillChangeAnimated` was getting called before `shouldChangeFromCamera`, so the `isUserInteraction` was always false.

This looks at the `MGLCameraChangeReason` bitmask with the updated Mapbox.framework library to make sure the reason bitmask includes any non-programmatic one (like a gesture or panning).

### DIFF
--- a/ios/RCTMGL/RCTMGLMapViewManager.m
+++ b/ios/RCTMGL/RCTMGLMapViewManager.m
@@ -477,7 +477,7 @@ RCT_EXPORT_METHOD(setCamera:(nonnull NSNumber*)reactTag
 
 - (void)mapView:(MGLMapView *)mapView regionWillChangeWithReason:(MGLCameraChangeReason)reason animated:(BOOL)animated
 {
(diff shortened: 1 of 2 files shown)


EricPKerr:
This works for iOS, but I'm not sure how to make the tests pass

nitaliano:
don't worry about the tests failing(since nothing is actually failing). Travis was turned on for all Mapbox repos I need to disable it since we are using circle

EricPKerr:
@nitaliano I've been using this PR in production on iOS for about a month now

nitaliano:
thanks 🍻 !

Issue #1130, opened 2018-03-21
Overlapping polygons at tile buffer when location dot enabled


kbauhausness:
On an Android device, darker/more opaque areas appear along what looks like the tile buffer. Looks like when the location dot is on a tile, it's the polygon on that tile that renders with a darker overlapping area. No need to zoom to reproduce.

Found using react-native-mapbox-gl version 6.1.0, which pulls in Android SDK version 5.4.1 (and iOS SDK version 3.7.3).

![image 4](https://user-images.githubusercontent.com/11279680/37742050-3dadaad2-2d21-11e8-8a87-29256d9db843.png)

nitaliano:
At first glance this is for sure related to the accuracy layer around the user location with the id `mapbox-location-accuracy-layer`, when I hide that layer when the map camera stops moving this artifact disappears.

kbauhausness:
Thanks @nitaliano. Note when troubleshooting, we also did a test in the MacOS test app from gl native using a proxy for the accuracy layer (a transparent circle) and some other semi-transparent polygons, but weren't able to replicate the issue that way. We used [this style](https://api.mapbox.com/styles/v1/danswick/cjf1eqwnj0xgu2rnosseym0dh.html?fresh=true&title=true&access_token=pk.eyJ1IjoiZGFuc3dpY2siLCJhIjoiY2l1dTUzcmgxMDJ0djJ0b2VhY2sxNXBiMyJ9.25Qs4HNEkHubd4_Awbd8Og#11.6/37.766868/-122.428816/-14). 

Also, may be worth trying to replicate this on iOS as I've heard that the overlapping polygons at tile buffer has been seen on iOS too.

nitaliano:
Ok, I'll try that style out in my test project on iOS and I'll let you know what happens.

jluxenberg-scoot:
@nitaliano @kbauhaus Any update you can share?

nitaliano:
I haven't been able to reproduce this issue yet on iOS.

From the react native standpoint while using a user location icon we won't be able to see this behavior on iOS since that is an annotation view, and Android is using sources/layers.

Some potential possible quick fixes could be
* We always hide the circle layer around the user location icon for the short term
* We can see if leaving the circle layer around the user location with a constant radius could help solve the problem at the tile boundaries, since in the android location layer plugin that get's updated based on GPS accuracy

nitaliano:
[...]

PR #1200, opened 2018-05-15
Release 6.1.2

nitaliano:
This is a short term fix to prevent weird glitches with the accuracy layer on top of fill layers, a more long term fix will be looked into once we update to the boba release

cc @kbauhaus


nitaliano:
I tested this walking around SF with my Pixel 2, and I tested different positions on my emulator. If there are still places where this get's reproduced please add the exact coordinate and I will test it out right away

kbauhausness:
Thanks @nitaliano. 

@jluxenberg-scoot and @garymccormick, please go ahead and test on your side & confirm if this fixes the overlapping polygons. If you're still seeing the issue in certain areas, as @nitaliano noted above it'd be useful to know the coordinates for those locations.

Related to https://github.com/mapbox/react-native-mapbox-gl/issues/1130

garymccormick:
@kbauhaus @nitaliano thanks for filing this. Do you have instructions for how we can test this from a branch?

nitaliano:
@garymccormick you can point to this branch in your package.json, or you could clone this repo and use the example application that comes bundled with it, and update the `ShowMap` example to point to the map style you would like to test with.

PR #1249, opened 2018-06-11
fixing bug with idle listener and region change events

cpetzel:
This fixes an issue where multiple calls to `setCamera` would break the regionDidChange callback because of the way this `reason` value was tracked.


nitaliano:
Thanks for the PR!

Issue #1238, opened 2018-06-05
[7.0.0] API Design


nitaliano:
For anyone that would like to leave feedback on what they would like to see in our upcoming major release please leave your feedback here.

edolix:
Wow, looking forward to see this release 👍 
I've found a strange bug described in #937 . 

The upcoming release deprecates the PointAnnotation? I've read something related around issues or gitter channel.
Thanks!

keenubee:
Do you plan to integrate routing?

nitaliano:
@edolix PointAnnotation is going to be removed, AnnotationViews no longer exist in the next version of the Android SDK. There are plans to drop an Annotation component that wraps sources/layers and gives a similar API to PointAnnotation

nitaliano:
@keenubee when you ask about routing are you talking about the nav SDK? That would be a separate react native SDK that we are debating if we should release or not

kristfal:
@nitaliano 

First of all, thanks a lot for your work on this wrapper. It's really good! Onto some wishes:

**1) Camera API**

_Current state:_
To fully use the camera API, a mix between `setCamera` and props on the `<MapView>` is required. 

Ex: During user location tracking, the android app takes into account zoomLevel on the `<MapView>`, meaning that any zoomLevel applied via `setCamera` doesn't propagate to the user location tracking state. There are also quite a few other issues related to the camera that do exist currently.

_Proposal:_
1) Go all in on a declarative prop-based camera API. 
2) Ensure that any combination of prop changes doesn't cause race conditions / ignored updates

Here is a proposal for camera related props: 

```
  animated: {Function | Bool} – can either be set to true/false, or passed a function that takes two arguments: (currentCamera, nextCamera) => which in turn should return Boolean to indicate camera transition should be animated
  animationDuration: {Function | Number} – if animated is true, sets the duration of the animation. Can also be passed function that takes two arguments: (currentCamera, nextCamera) => which in turn should return a Number to indicate transition duration
  animationType: {Function | String} – If string, one of 'FlyTo' or 'JumpTo'. Can also be passed function that takes two arguments: (currentCamera, nextCamera) => which in turn should return a String to indicate transition type
  center: {Array} – camera center 
  zoom: {Number} – camera zoom
  bearing: {Number} – camera bearing
  pitch: {Number} – camera pitch
  bounds: {Array} – camera bounds. Setting this after setting center or zoom will override center and zoom and vice versa
[...]

PR #1252, opened 2018-06-14
Make the MapView Native class extendable on Android

cpetzel:
We want to make some core changes to how MapView instance, and would like to create a subclass of RCTMGLMapView.java. We made these changes on our fork and it works great!

The gist of these changes is making a few private members/methods protected so the child class can interact with them. 

The main motivating factor for these changes was to make the location engine stay active regardless whether `showUserLocation` was enabled or not.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
@@ -8,6 +8,7 @@ import android.view.View;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.common.MapBuilder;
+import com.facebook.react.common.MapBuilder.Builder;
 import com.facebook.react.uimanager.LayoutShadowNode;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.annotations.ReactProp;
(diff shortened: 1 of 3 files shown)


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapViewManager.java
```diff
@@ -245,20 +246,27 @@ public void setUserLocationVerticalAlignment(RCTMGLMapView mapView, int userLoca
...
+            .put("queryRenderedFeaturesAtPoint", METHOD_QUERY_FEATURES_POINT)
+            .put("queryRenderedFeaturesInRect", METHOD_QUERY_FEATURES_RECT)
+            .put("getVisibleBounds", METHOD_VISIBLE_BOUNDS)
+            .put("getPointInView", METHOD_GET_POINT_IN_VIEW)
+            .put("getCoordinateFromView", METHOD_GET_COORDINATE_FROM_VIEW)
+            .put("takeSnap", METHOD_TAKE_SNAP)
+            .put("getZoom", METHOD_GET_ZOOM)
+            .put("getCenter", METHOD_GET_CENTER);
+
+    }
```
cpetzel, line 264: This will allow us to add our own custom methods on the child class. The child can simply override this method and add his method declarations to this collection.

nitaliano:
@cpetzel I can merge this in, but in `7.0.0` the location will be decoupled from the Map and will exist under `MapboxGl.locationManager`

cpetzel:
Thats great! If things are going to drastically change with locationManager in 7, then there is no need to merge this in. We will maintain our fork until 7.0. 

Is there a branch where I can take a peek at the changes? Any idea on the timeline? 

Thanks!

nitaliano:
This is what I was thinking for 7 https://github.com/mapbox/react-native-mapbox-gl/issues/1238#issuecomment-397435490, the userlocation marker will also not be coupled to the map anymore

Issue #1209, opened 2018-05-21
No way to track panning/zooming gestures 


amitassaraf:
There is currently no way to my knowledge to track a pan / zoom gesture starting when it begins and all the way to when it ends.

Example: I want to hide a UI element while the user is dragging / panning around the map.

KevinHu2014:
@amitassaraf 
Maybe you can combine using `onRegionDidChange ` and `getZoom()`  together. 
When the `onRegionDidChange ` was triggered , use  `getZoom()` to see whether the zoom level has be changed.

sandroferrari:
The `onRegionDidChange` gives you a GeoJson feature which contains `zoomLevel` in their properties, you can proceed like this :
`onRegionDidChange={ (args) => {
    let zoom = args.properties.zoomLevel; 
 } }`

amitassaraf:
`onRegionDidChange` is called only once the user has stopped dragging. I want to actually track the "onTouchDown" and "onTouchUp" of the user on the map

KevinHu2014:
There are also `onRegionWillChange ` and `onRegionIsChanging` that you can use.

sandroferrari:
Pay attention to this #980

chrusart:
@amitassaraf  check my PR if it will help.

PR #1310, opened 2018-07-26
[Android] Region changed events improvements

chrusart:
I came to this PR from #1209 which I also need.

The thing is that ```onRegionWillChange``` didn't work always which this PR fixes.

Problem was that before ```onMapChange``` ```REGION_WILL_CHANGE``` event triggers, ```onTouchEvent``` with ```ACTION_MOVE``` is triggered multiple times, incrementing ```mChangeDelimiterSuppressionDepth``` and then ```REGION_WILL_CHANGE``` is suppressed before even first time sent to js.

I found no reason to suppress ```onRegionWillChange``` with ```animated``` flag, and I removed this check, actually they fire correctly when user do 'flick' gesture. It's triggered when sdk do scroll momentum after flick, or programmatically will animate to new region.

Actually my problem was to have own debouncing on didChange/willChange events and cancel _.debounce when user interacts with map (not talking about delays, yes, I found them without a word in docs :) ). Not sure I can do what I want since didChange debouncing will debounce on didChange event of course, but when user move, release finger and quickly want to swipe more, debounce will trigger didChange event, but I don't want to trigger it until user finish swiping. So if taking up finger, putting back and continue swiping I want to cancel didChange debounce to not trigger.

Now on ```REGION_WILL_CHANGE``` event (not animated, animated is after didChange) I can cancel my debouncing on didChange.

Other thing is what I didn't fix here (will put in issue) is that ```isUserInteraction``` is 
not set correctly, but have no idea how to properly fix it.
EDIT: PR updated with fix for it.
```onCameraMoveStarted``` actually sets correct reason which is used to set ```isUserInteraction``` but it fires after ```onMapChange``` ```REGION_WILL_CHANGE``` event (or other events).
With this PR I have log:
```
ACTION_MOVE
REGION_WILL_CHANGE (isUserInteraction() - false)
CameraMoveStartedListener true
REGION_WILL_CHANGE ANIMATED <- this is user momentum scroll animation
CameraMoveStartedListener true (true - mCameraChangeTracker.isUserInteraction())
// here I get didChange event in js, fetch some data,
// update store with new bounds, and fitBounds is triggered (in this case unnecessary)
REGION_WILL_CHANGE ANIMATED <- this is after fitBounds programmatically
CameraMoveStartedListener false (false - mCameraChangeTracker.isUserInteraction())
```
so ```REGION_WILL_CHANGE[_ANIMATED]``` are sent with incorrect ```isUserInteraction``` flag.
This problem is probably connected to #980 (iOS)

I hope I've tested and described everything correctly.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -1,22 +1,18 @@
 package com.mapbox.rctmgl.components.mapview;
 
-import android.app.Activity;
 import android.content.Context;
-import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.PointF;
 import android.graphics.RectF;
 import android.location.Location;
 import android.os.Handler;
 import android.support.annotation.NonNull;
-import android.text.LoginFilter;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.MotionEvent;
 
-import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
@@ -39,9 +35,7 @@ import com.mapbox.mapboxsdk.maps.MapboxMapOptions;
 import com.mapbox.mapboxsdk.maps.OnMapReadyCallback;
 import com.mapbox.mapboxsdk.maps.UiSettings;
 import com.mapbox.mapboxsdk.plugins.localization.LocalizationPlugin;
-import com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerMode;
 import com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerPlugin;
-import com.mapbox.mapboxsdk.storage.FileSource;
 import com.mapbox.mapboxsdk.style.layers.Layer;
 import com.mapbox.mapboxsdk.style.layers.Property;
 import com.mapbox.mapboxsdk.style.layers.PropertyFactory;
@@ -380,12 +374,14 @@ public class RCTMGLMapView extends MapView implements
             mMap.moveCamera(CameraUpdateFactory.newCameraPosition(buildCamera()), new MapboxMap.CancelableCallback() {
                 @Override
                 public void onCancel() {
-                    sendRegionChangeEvent(false);
+                    sendRegionDidChangeEvent(false);
+                    mCameraChangeTracker.setReason(-1);
                 }
 
                 @Override
                 public void onFinish() {
-                    sendRegionChangeEvent(false);
+                    sendRegionDidChangeEvent(false);
+                    mCameraChangeTracker.setReason(-1);
                 }
             });
         }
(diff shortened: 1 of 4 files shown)


chrusart:
From what I see in code events are mixed from ```onMapChange``` and camera listeners.
I will propose some more changes in this PR in few hours which resolve some issues (mentioned above) and simplify code.

chrusart:
I made new changes:
- i tried to base all region changes events handling only on camera listeners (almost succeded), it's a nice separation from specific for map events and user interaction (camera/region) events.
- removed ```mChangeDelimiterSuppressionDepth``` as not needed now
- removed time and animated condition to send DID_CHANGE event as it should be a bridge and hiding/suppressing events shouldn't be a case, maybe it was there before debouncing or sending DID_CHANGE on fling, anyway, I'm sure such logic can be done by developer and events shouldn't be suppressed (yes, i did some suppressing as a workaround to fling stuff, but a small one :), read below).
- I had to use some flag to distinguish that it's a fling and I found that only fling triggers onCameraIdle between WLL_CHANGE_ANIMATED and DID_CHANGE_ANIMATED events so I added isAnimating.
- isUserInteraction and isAnimated should be computed from ```reason``` given in ```onCameraMoveStarted``` and used in ```onCameraIdle``` to send DID_CHANGE event

One workaround now is to set isAnimating, then we know it were a fling animation events.
If we have ```addOnFlingListener``` we shouldn't send WILL_CHANGE and DID_CHANGE events during region change when fling starts. Then we could remove this workaround, but probably some native SDK changes needed.

I put some questions for maintainers in code as ```reason``` comes from native SDK.
When we get ```reason == 2```?
Can we have SDK ```reason```, but not animated?

On user gesture we have ```reason == 1```, when we do ```fitBounds``` we have ```reason == 3``` (even if we set delay to 0)

JS - sending WILL_CHANGE/DID_CHANGE events immediately if debounce is set to 0 (not on next frame)

Will try to do some events whitelist to not send through bridge events we don't listen to (while waiting for maintainers response). There is a LOT of events (frame, region is changing) that are send now, but if not listen to, can be filtered in ```shouldDropEvent```.

If this PR will break something I wasn't aware, will update to make all work again.

chrusart:
Now JS side sends list of events that it handling in props (on....) that native code will not send unnecessary events through bridge.

```moveCamera``` listener ```onCancel``` and ```onFinish``` were sending REGION_DID_CHANGE events with ```animated``` hardcoded false. I guess ```animated``` should be taken from ```mCameraChangeTracker``` set when ```onCameraMoveStarted``` is triggered. Now it takes ```animated``` from ```mCameraChangeTracker``` in this case.

chrusart:
Ok, I can check mapbox-gl-native for ```reason``` and implementation.

chrusart:
Was on vacations. ```reason``` with 2 value didn't find anywhere in native code.
Could someone from maintainers give some feedback on this PR?
Is it a good direction? @nitaliano

kristfal:
@nitaliano we've been using this in production for a while without any issues. I'd suggest we merge this to master so iOS and Android is aligned. Any objections?

Daavidaviid:
Some news ?
Region changed events really are broken on android. They work the first time you launch the app. But as soon as your reload it it's over.

kristfal:
@chrusart Thanks alot for your contribution. I'm up for merging this one in. No word from the main maintainer, but it looks solid and we've been using it for a while without issue.

Can you resolve conflicts and ping me for merge? Thanks!

kristfal:
These changes are now in master.

PR #1320, opened 2018-08-01
Avoir crash when URL is null (Vector/Shape/Raster Source)

Gp2mv3:
Currently, if the url of a source is null or undefined, the app crashes (On Android at least).
This ensure that the url is defined and not null before trying to render the native component that crashes.

### DIFF
--- a/javascript/components/RasterSource.js
+++ b/javascript/components/RasterSource.js
@@ -73,6 +73,9 @@ class RasterSource extends React.Component {
       tms: this.props.tms,
       attribution: this.props.attribution,
     };
+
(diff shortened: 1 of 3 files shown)


nitaliano:
This should be reverted. A ShapeSource also allows a shape to be passed down so it should not return null if a url is not provided. You can also use a VectorSource without a url by hitting the composite source. I would recommend testing every example in the example app before merging

Gp2mv3:
I reverted it. Sorry about that.

nitaliano:
No need to be sorry it's all a learning experience!

Issue #1250, opened 2018-06-13
JS required image not working with ImageSource in release mode


beisert1:
I am getting a blank white screen when loading a map running from a signed APK build on Android. I only see the mapbox info icon.

It works fine on debug for Android on all my devices, it also works if I set debuggable to true in the gradle file but obviously I am not able to send a debuggable build to google play store.

Works fine on release build for iOS. I am also not seeing any errors or anything in the logcat when I debug.

Any help would be greatly appreciated.

nitaliano:
how are you generating your APK? Can you post your `app/build.gradle`?

beisert1:
@nitaliano I followed the documentation here to generate the APK. [https://facebook.github.io/react-native/docs/signed-apk-android.html#content](https://facebook.github.io/react-native/docs/signed-apk-android.html#content)

Here is the app/build.gradle

```
apply plugin: "com.android.application"

import com.android.build.OutputFile

project.ext.react = [
    entryFile: "index.js"
]

apply from: "../../node_modules/react-native/react.gradle"

def enableSeparateBuildPerCPUArchitecture = false

def enableProguardInReleaseBuilds = false

android {
    compileSdkVersion 26
    buildToolsVersion "26.0.1"
    defaultConfig {
        applicationId "com.imts18"
        minSdkVersion 16
        targetSdkVersion 26
        versionCode 1
        versionName "1.0"
        ndk {
            abiFilters "armeabi-v7a", "x86"
        }
    }
    signingConfigs {
        release {
            if (project.hasProperty('MYAPP_RELEASE_STORE_FILE')) {
                storeFile file(MYAPP_RELEASE_STORE_FILE)
                storePassword MYAPP_RELEASE_STORE_PASSWORD
                keyAlias MYAPP_RELEASE_KEY_ALIAS
                keyPassword MYAPP_RELEASE_KEY_PASSWORD
            }
        }
    }
    splits {
        abi {
            reset()
            enable enableSeparateBuildPerCPUArchitecture
            universalApk false  // If true, also generate a universal APK
            include "armeabi-v7a", "x86"
        }
    }
    buildTypes {
        release {
            signingConfig signingConfigs.release
            minifyEnabled enableProguardInReleaseBuilds
            proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
        }
    }
    // applicationVariants are e.g. debug, release
    applicationVariants.all { variant ->
[...]

PR #1331, opened 2018-08-11
Fix ImageSource not support require local file on Android

rogerhokp:
Refer: https://github.com/mapbox/react-native-mapbox-gl/issues/1250

Fix ImageSource`url` not support require local file on Android


matthieucan:
@rogerhokp thanks!
I can confirm this patch works and fixes #1250
Is there a blocker to get it merged?

pinpong:
Why this PR is not merged yet?

nitaliano:
@rogerhokp thanks for the PR!

janweigel:
We're still having problems to load images with require if we're in release mode or not connected to the local packager in debug mode.
Error message: "Failed to load source xyz"

If using an image with require(...) for MapboxGL.ImageSource it still throws this error if the url is not a remote URL (https://...) or loaded from the local filesystem (file://...). 
It should be possible to load images provided with the app bundle with require. 

Is there any workaround to reliably load images from the app bundle with require?
Thanks!

PR #861, opened 2017-12-12
[Android]requestDisallowInterceptTouchEvent in touchEvent

Tennen:
mapview in a scrollview is hard to scroll
disallow intercept touch event by parent view can fix this issue.
I don't see any negative effects for now.

Issue #856, opened 2017-12-11
scroll event propagation


Tennen:
I have a MapView in a vertical ScrollView, when I scroll the map, and the scroll action has a vertical speed, the ScrollView will start to scroll. It seems to be a problem of scroll event propagation on android.
the version I use now is 6.0.1, but I didn't face this problem in v5.2.1.
I tried using PanResponder but it doesn't work.
does any one know how to solve it?
```
<KeyboardAwareScrollView>  
    <MapboxGL.MapView  
          rotateEnabled
          scrollEnabled
          zoomEnabled
          logoEnabled={false}
          compassEnabled
          ref={(mapRef) => {
            this.mapRef = mapRef;
          }}
          style={styles.map}
          centerCoordinate={this.state.center}
          zoomLevel={this.state.zoom}
          showUserLocation={false}
          userTrackingMode={this.state.userTrackingMode}
          onDidFinishLoadingMap={this.fitBounds}
        >
</KeyboardAwareScrollView>
```

I override the onTouchEvent of RCTMGLMapView and requestDisallowInterceptTouchEvent, it works
I will take a PR

Tennen:
fixed by this [pr](https://github.com/mapbox/react-native-mapbox-gl/pull/861), close.

fnusneha:
This PR causes a different issue though. Since the onTouchEvent() blindly requests to disallow intercept, even when the map is completely static, scroll events get captured in the map, and parent doesn't scroll. Outside the map, the parent container scrolls normally.

Could this be changed to a conditional disablement, where if the map is static, this does not happen?

chfight:
I added refreshControl in ScrollView. RefreshControl is appeared when I pull the map, and the map doesn't move. 
This problem only happened in android. 
Can you help me?

PR #1342, opened 2018-08-23
Don't intercept touch event if scroll is disabled

fnusneha:
We have a MapView ( with scroll disabled) in a vertical ScrollView. If the user touches on the mapView to scroll the parent view then nothing happens. If the users touches anywhere else in the scroll view, then it works fine.

Reason
Inside Mapview.onTouchEvent, we are stopping the propagation of the event to the parent view. As a result, nothing happen when we try to scroll the scroll view while touching the mapView. In order to temporary fix this, we added another condition to stop propagating the event . Only when scroll is enabled in the mapView, we disable propagating the event.

This-https://github.com/mapbox/react-native-mapbox-gl/issues/856  was an issue raised long back which was fixed by this PR-https://github.com/mapbox/react-native-mapbox-gl/pull/861. However, This PR causes a different issue though. Since the onTouchEvent() blindly requests to disallow intercept, even when the map is completely static, scroll events get captured in the map, and parent doesn't scroll. Outside the map, the parent container scrolls normally.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -542,7 +542,7 @@ public class RCTMGLMapView extends MapView implements
             mChangeDelimiterSuppressionDepth = 0;
         }
 
-        if (result) {
+        if (result && mScrollEnabled) {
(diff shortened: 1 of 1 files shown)


nitaliano (approved):
LGTM, I'll test this out today

nitaliano:
tested it out works for me, I'll publish a new package tomorrow

nitaliano:
@fnusneha 6.1.3 is published

PR #1392, opened 2018-10-11
NullpointerException fix in RCTMGLMapView, when the camera position h…

fnusneha:
We were getting this crash in our weatherbug application-
Stack traces:
 
Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'double com.mapbox.mapboxsdk.geometry.LatLng.getLatitude()' on a null object reference
 
com.mapbox.rctmgl.components.mapview.RCTMGLMapView.makeRegionPayload (Unknown Source:10)
com.mapbox.rctmgl.components.mapview.RCTMGLMapView.sendRegionChangeEvent (Unknown Source:2)
com.mapbox.rctmgl.components.mapview.RCTMGLMapView.access$700 (Unknown Source)
com.mapbox.rctmgl.components.mapview.RCTMGLMapView$3.onCameraIdle (Unknown Source:58)
com.mapbox.mapboxsdk.maps.CameraChangeDispatcher$4.run (Unknown Source:66)
android.os.Handler.handleCallback (Handler.java:789)
android.os.Handler.dispatchMessage (Handler.java:98)
android.os.Looper.loop (Looper.java:164)
android.app.ActivityThread.main (ActivityThread.java:6938)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:327)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1374)

So, added null check for this.

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
@@ -1258,6 +1258,9 @@ public class RCTMGLMapView extends MapView implements
 
     private WritableMap makeRegionPayload(boolean isAnimated) {
         CameraPosition position = mMap.getCameraPosition();
+        if (position == null || position.target == null) {
+            return new WritableNativeMap();
+        }
         LatLng latLng = new LatLng(position.target.getLatitude(), position.target.getLongitude());
 
         WritableMap properties = new WritableNativeMap();


android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
```diff
@@ -1258,6 +1258,9 @@ private void updateLocationLayer() {
 
     private WritableMap makeRegionPayload(boolean isAnimated) {
         CameraPosition position = mMap.getCameraPosition();
+        if (position == null || position.target == null) {
+            return new WritableNativeMap();
```
zugaldia, line 1262: I checked and in fact `GeoJSONUtils.toPointFeature()` doesn't support a `null` value for the location, neither does this method. While a `null` check seems the a good next step, two questions come to mind:
1. Do you expect any side issues from returning an empty map here?
2. Could `position` being `null` here indicative of a larger issue, or is it expected to have a `null` value in some circumstances?

zugaldia (commented):
Thank you @fnusneha for the contribution and sorry to hear you're hitting that crash. Just dropped a short note on your change. Let me know your thoughts when you have a chance.

android/rctmgl/src/main/java/com/mapbox/rctmgl/components/mapview/RCTMGLMapView.java
fnusneha: This empty map is being returned as a payload for MapChangeEvent. At least in Mapbox code itself, I’m not seeing any usage of this event or its payload.

Since this is the only crash log we have, it is not clear exactly at what step this happens. However, I assume onCameraIdle would be called at some step where camera is not active (due to not having location or activity closing?). So it might be possible that it has no location set.

zugaldia: Gotcha @fnusneha - thanks for the additional info. Looks like this is the best approach we can take short term while we research further if there's a deeper cause for this issue. I'll merge this PR now, let me know if you manage to get any new information. Thanks again for your contribution!

fnusneha: Thanks @zugaldia

PR #1455, opened 2018-12-21
Get visible bounds in view coordinates

ibesora:
Adds the method *getVisibleBoundsInViewCoordinates* to *MapboxGL.MapView*.
Currently, in order to get features visible in the map you have to call *getVisibleBounds* which returns the coordinate bounds visible in the user's viewport, convert those coordinates to view coordinates via *getPointInView* and then use *queryRenderedFeaturesInRect* with those coordinates.
*getVisibleBoundsInViewCoordinates* aims to make this process easier by removing the conversion step between world coordinates and view coordinates.

### DIFF
--- a/docs/MapView.md
+++ b/docs/MapView.md
@@ -93,6 +93,22 @@ const visibleBounds = await this._map.getVisibleBounds();
 ```
 
 
+#### getVisibleBoundsInViewCoordinates()
+
+The user's viewport bounds in the given view's coordinate system.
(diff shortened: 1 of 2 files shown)


kristfal:
Hey, sorry for not getting back to you. I propose we keep this one out to keep the surface area of the API as small as possible. Imo, helpers like this should be added on an app to app basis rather than being part of the core library.

PR #1456, opened 2018-12-21
Small documentation fix

ibesora:
Modifies the MapView documentation to better reflect that *queryRenderedFeaturesAtPoint* and *queryRenderedFeaturesInRect* are async methods and they should be called with await.


kristfal:
Looks good.

PR #1485, opened 2019-01-23
update NSNumber prop attributes from assign -> strong

### LINKED ISSUE #1450: EXC_BAD_ACCESS: hash > retain > style > Attempted to dereference garbage pointer 0xe.
After our latest release we started getting hammered by EXC_BAD_ACCESS fetal errors and NSInvalidArgument exceptions. We have been unable to reproduce the issue locally, and are only seeing crashes with iOS. We are using Sentry for our production error reporting. 

Environment: 
"@mapbox/react-native-mapbox-gl": "^6.1.2",
 "react-native": "0.57.8",

Stack trace: 
screen shot 2018-12-21 at 10 24 03 am

L-Yeiser:
Proposed fix [for](https://github.com/mapbox/react-native-mapbox-gl/issues/1450)

### DIFF
--- a/ios/RCTMGL/CameraStop.h
+++ b/ios/RCTMGL/CameraStop.h
@@ -10,14 +10,14 @@
 
 @interface CameraStop : NSObject
 
-@property (nonatomic, assign) NSNumber *pitch;
-@property (nonatomic, assign) NSNumber *heading;
-@property (nonatomic, assign) NSNumber *zoom;
-@property (nonatomic, assign) NSNumber *boundsPaddingLeft;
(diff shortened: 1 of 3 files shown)


kristfal:
@l-yeiser can you confirm if this fix does indeed solve the issue in production? (Ex: push your fork to production and check for crashes)

L-Yeiser:
The fork was released to production two weeks ago and we have not seen that crash on that release.

orta:
Can confirm, an NSObject subclass like these should never be `assign`

nitaliano:
I totally agree with this thanks for the PR! 😄

PR #1491, opened 2019-01-31
Fix Android UIManager deprecation warning for RN 0.58

marudy:


kristfal:
@marudy is this backwards compatible? If so, how far back?

goodryanboy:
Is there any workaround here?

kristfal:
This will break RN 0.57 or lower. Can you add a fallback for those versions?

kristfal:
Closing this. Will accept a PR with a fallback for lower versions of RN.

alexiri:
Hi @kristfal, you rejected this PR but you introduced the change anyway as part of 88f9f7f9279031417a3964c5567a2b77577fbf99. Was this intentional?

kristfal:
Hey, that was not intentional. I think it should have been reversed.

PR #1504, opened 2019-02-06
Podspec fix for people using CocoaPods for distribution and not NPM

orta:
Typically during the install process `npm` calls `node ./scripts/download-mapbox-gl-native-ios-if-on-mac.js 3.7.8` to download the MapBox SDK: [here](https://github.com/mapbox/react-native-mapbox-gl/blob/master/package.json#L26)

However, when you install this library via CocoaPods (when [separating your JS / native code](http://artsy.github.io/blog/2018/04/17/making-a-components-pod/) ) then this script doesn't run. This allows CocoaPods consumers to also get a copy of the framework.

### DIFF
--- a/react-native-mapbox-gl.podspec
+++ b/react-native-mapbox-gl.podspec
@@ -14,5 +14,7 @@ Pod::Spec.new do |s|
   s.source_files	= "ios/RCTMGL/**/*.{h,m}"
 
   s.vendored_frameworks = 'ios/Mapbox.framework'
+  s.preserve_paths = ['scripts']
+
   s.dependency 'React'
 end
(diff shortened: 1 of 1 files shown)


nitaliano:
@orta what will happen if I pull in the native mapbox ios sdk thru my podfile?

orta:
Hrm, good Q, let's see how that works - might be able to close this

orta:
Looking good, I get framework conflicts. Perhaps this PR can be switched to documentation on the README.

nitaliano:
@orta thanks for checking it out, I with updating the documentation on this PR and we can get it merged in.

orta:
Happy to see where this goes instead: https://github.com/nitaliano/react-native-mapbox-gl/pull/1533

PR #1505, opened 2019-02-09
Upgrade mapbox-android-sdk to 5.5.3

alexiri:
Versions older than 5.5.2 suffered from a [bug](https://github.com/mapbox/mapbox-gl-native/pull/11583) that prevented Android emulators from showing the map when using hardware acceleration. This upgrades to SDK 5.5.3, the last of the 5 series, which doesn't have this issue.


kristfal:
This is already in master via an alternative PR. Thanks for your contribution.

Issue #1189, opened 2018-05-08
MapView onPress event not firing @ 6.1.1 & react-native@0.55.3


jussirantala:
Hi,

I'm using react-native@0.55.3 and react-native-mapbox-gl@6.1.1 in Genymotion Android 7.1.

It seems that onPress event of MapView is not getting fired. Nothing happens when I click on the map but the map is otherwise working fine.

I think this started to happen after updating react-native and react-native-mapbox-gl to latest versions.

Is anyone else experiencing this bug?

```
<MapboxGL.MapView
          logoEnabled={false}
          ref={(map) => {this.mapRef = map}}
          styleURL={this.mapStyleURI}
          compassEnabled={false}
          zoomLevel={2}
          centerCoordinate={[southWestLatLng.longitude, southWestLatLng.latitude]}
          style={styles.mapView}
          attributionEnabled={false}
          onRegionDidChange={this.onMapRegionChange}
          onPress={() => console.log('TEST')}>

          <MapboxGL.ShapeSource id='markerSource' shape={markerCoordinates}>
            <MapboxGL.SymbolLayer
              id='markerSymbols'
              style={mapStyles.defaultMarker} />
          </MapboxGL.ShapeSource>

</MapboxGL.MapView>
```

jussirantala:
Seems to be working on a production build in real device with Android 7.

cvietor:
same problem here after updating from 6.1.0 to 6.1.1. Events stopped working (onPress, onLongPress, onRegionDidChange) while debugging on a real device (android 7 phone)

cvietor:
I'm having a similar issue (debug mode on real android 7.0 device). onPress, onLongPress, onRegionDidChange events (these are the ones tested by me) stopped working after updating to the latest version of react-native.

What i found so far: When freshly installed my app in debug mode, events are working. As soon as i attach the debugger, the events stop working. They will work again, if you detach the debugger, kill and restart your app.

jcsena:
The same problem with the ShapeSource ```onPress```

janimalin:
Had same issue with 6.1.2-beta2 / Android 7 / react-native 0.55 (expo) all onPress events failed to trigger. Downgrading to at least 6.0.3 fixed it.

nitaliano:
I'm not able to reproduce this issue at all, would someone be able to provide steps and if you're using the debugger or not? We just shipped an app with `6.1.2-beta2` and all the presses are working.

[...]

PR #1508, opened 2019-02-14
Added EventDispatcher to dispatch events

lLenn:
I'm using expo and had issues with events not triggering because they don't implement ReactApplication.
I've changed the code so that it's based off of standard React-Native components and in my opinion this is the proper way to dispatch events.
This triggers the events properly with expo, but also for normal react-native projects.
 #1189

### DIFF
--- a/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/AbstractEvent.java
+++ b/android/rctmgl/src/main/java/com/mapbox/rctmgl/components/AbstractEvent.java
@@ -1,0 +1,28 @@
+package com.mapbox.rctmgl.components;
+
+import com.facebook.react.bridge.WritableMap;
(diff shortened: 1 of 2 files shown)


kristfal:
@lLenn, thanks for your contribution, this looks good. As you may have noticed, we have a workaround in place for reconnecting when the JS side is reloaded. I'm not very well versed on the Android side, so can you just verify/confirm that reloading the JS bundle reconnects properly? Thanks

kristfal:
Hey. This looks good. Just to double check: Does it reconnect properly when the JS side is reloaded?

lLenn:
@kristfal, yes, that was the problem I encountered aswell, but your patch only worked for normal rn projects. Now it works for expo projects aswell.

kristfal:
Very good. Lgtm.

PR #1522, opened 2019-03-04
updated example

onitzschke:


onitzschke:
@kristfal  can you please review and fix the issue with the accesstoken?

kristfal:
Thanks for contributing, this looks great. What is the issue with accesstoken? Right now, you'll need to create a file called accesstoken with your mapbox at.

onitzschke:
It's not possible to get the accesstoken by [getDirections()](https://github.com/onitzschke/react-native-mapbox-gl/blob/master/example/src/components/DriveTheLine.js#L83) because make_service.js throw an error.

onitzschke:
@kristfal the issue with accesstoken is fixed with 5a7e03e.
The other problem ist that map is not show with MapboxGL style url's but worked with our tileserver.
Any idea?

onitzschke:
Sorry, my fault. Something is wrong with my access token.

onitzschke:
@kristfal i've tested on both platform and all is working fine.
I think this is a good base to get v7 ready.

kristfal:
I agree. Let's get this one in.

PR #1526, opened 2019-03-05
Support for generating type definitions for prototypes

orta:
This isn't a complete map of the library, just the prototypes - next up there needs to be a mapping of the component to their interfaces. Looks like:

```ts
/* eslint-disable */
// DO NOT MODIFY
// THIS FILE IS AUTOGENERATED

export interface ConstantProps {
  styletype: string;
  payload: {
    value: any;
  };
}

export interface StyleFunctionProps {
  styletype: string;
  payload: {
    fn: string;
    attributeName: string;
    stops: any[];
    mode: any;
  };
}

export interface TransitionProps {
  duration: number;
  delay: number;
}

export type TranslationProps = { x: number; y: number } | number[];

export interface fillLayerStyleProps {
  /**
   * Whether this layer is displayed.
   */
  visibility: any;

  /**
   * Whether or not the fill should be antialiased.
   */
  fillAntialias: TransitionProps | string | StyleFunctionProps;

  /**
   * The opacity of the entire fill layer. In contrast to the `fillColor`, this value will also affect the 1px stroke around the fill, if the stroke is used.
   */
  fillOpacity: TransitionProps | number | StyleFunctionProps;

  /**
   * The transition affecting any changes to this layer’s fillOpacity property.
   */
  fillOpacityTransition: TransitionProps;

  /**
   * The color of the filled part of this layer. This color can be specified as `rgba` with an alpha component and the color's opacity will not affect the opacity of the 1px stroke, if it is used.
   *
   * @disabledBy fillPattern
   */
  fillColor: TransitionProps | string | StyleFunctionProps;

  /**
   * The transition affecting any changes to this layer’s fillColor property.
   */
  fillColorTransition: TransitionProps;

  /**
   * The outline color of the fill. Matches the value of `fillColor` if unspecified.
   *
   * @disabledBy fillPattern
   */
  fillOutlineColor: TransitionProps | string | StyleFunctionProps;

```

and so on.


kristfal:
Thanks a lot for your contribution. Merging in! Will be happy to get more of em 👍

PR #1527, opened 2019-03-06
fallback for case when children is already an array

### LINKED ISSUE #1524: TypeError: null is not an object (evaluating 'foundChildren.filter')
```
 TypeError: null is not an object (evaluating 'foundChildren.filter')

     This error is located at:
      in ShapeSource (at createAnimatedComponent.js:153)
     in AnimatedComponent (at PulseCircleLayer.js:120)
     in PulseCircleLayer (at createAnimatedComponent.js:153)
     in AnimatedComponent (at DriveTheLine.js:126)
     in RCTMGLAndroidTextureMapView (at MapView.js:890)
```

sethvincent:
This is a small fix that ensures `foundChildren` won't be null when the filter method is called in the case where `children` is already an array.

Closes #1524

### DIFF
--- a/javascript/utils/index.js
+++ b/javascript/utils/index.js
@@ -69,6 +69,8 @@ export function cloneReactChildrenWithProps(children, propsToAdd = {}) {
 
   if (!Array.isArray(children)) {
     foundChildren = [children];
+  } else {
+    foundChildren = children
   }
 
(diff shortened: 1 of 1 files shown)


onitzschke:
Cool but this was also fixed with PR https://github.com/mapbox/react-native-mapbox-gl/pull/1522 ^^

PR #1544, opened 2019-03-18
RCTMGLImageSource.m fixed to be working with local files iOS

gigby:
RCTMGLImageSource.m fixed to be working with local files iOS. Now ImageSource is displaying only HTTP(HTTPS) URLs. But in docs it said that this component may work with `absolute file URL, or local file URL to the source image.` ([link](https://github.com/nitaliano/react-native-mapbox-gl/blob/master/docs/ImageSource.md))

### DIFF
--- a/ios/RCTMGL/RCTMGLImageSource.m
+++ b/ios/RCTMGL/RCTMGLImageSource.m
@@ -23,9 +23,20 @@
 
 - (MGLSource *)makeSource
 {
+    NSURL *myURL;
+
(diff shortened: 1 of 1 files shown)


kristfal:
Just to double check, this works on Android, right?

gigby:
@kristfal 
Yes, it works on Android as described in docs. Urls like "Http / Https" or "file: //" on Android work correctly

kristfal:
Ok, great. Merging. Thanks for your contribution @gigby!

PR #1545, opened 2019-03-19
Add Typescript declarations

RyPope:
### Description
I based these typings mostly off of the documentation. I knew some stuff was missing so I went in and added it, such as the access token and telemetry functions. 

Some of it can be improved - I'd really like to find a way to enforce [lat, lng] and [top, bottom, left, right] types so if anyone has an idea let me know.

### Testing
I manually added an index.d.ts to the node_modules folder of a project I'm working on and did some sanity to ensure that the typing is being picked up and it is.


kristfal:
Hey,

Thanks a lot for the contribution. FYI, the docs are missing some methods for OfflineManager. Would be able to check the code and build the interfaces from there (alternatively I can merge as is and then we can follow up). 

For clarity, these methods all lack docs: https://github.com/nitaliano/react-native-mapbox-gl/blob/master/javascript/modules/offline/OfflinePack.js

RyPope:
Thanks for the feedback. I added the OfflinePack definitions. I was having some trouble figuring out how to have multiple namespace declarations in the same index.d.ts but it works if I add it to the subdirectory of the module, not sure which is more desirable - having everything in a single index.d.ts at the base directory or having index.d.ts spread out through the directories.

PR #1551, opened 2019-03-20
Add region is changing event payload

kristfal:
This brings the `mapView` event `onRegionIsChanging` in line with the other mapView change events. The event will now send the following payload:

```js
  coordinates: <Array<number>>,
  zoomLevel: <number>,
  heading: <number>,
  pitch: <number>,
  animated: <boolean>, // always false
  isUserInteraction: <boolean>
  visibleBounds: <Array<number>>
```


mattijsf:
Nice this might allow us to integrate a custom compass widget on top of the map. The previous onRegion[Will/Did]Change was not "real-time" enough.

kristfal:
Yes, we may need to throttle it slightly to not flood the bridge, in particular on iOS, but it is a step in the right direction.

About controls: I think we should stop using all native controls and native custom components (attribution, compass, user location marker, user location camera behaviour) and control these from the RN side. That gives us a fully uniform behaviour across iOS and Android.

I've also been looking at how we manage styles and camera changes, and I think we should slim down the surface area greatly. My main concern is that if we start doing these changes now, we'll make the work of getting @nitaliano's 7.0.0 changes in more difficult.

PR #1556, opened 2019-03-21
Dynamic version for Android Support Library and OKHTTP

sfratini:
Hey, so I was having problems building with master, and it was because I was compiling with version 27 and Mapbox was forcing version 28 of the Support Library. 

You get issues like this: https://stackoverflow.com/questions/49280632/error9-5-error-resource-androidattr-dialogcornerradius-not-found 

So I had to force my compileSdkVersion to 28. 

Eventually I made the supportLibVersion dynamic and also the okhttp version since some people were having issues. Bumped the compile version to 28 since it makes sense to have the same as the support library.

### DIFF
--- a/android/install.md
+++ b/android/install.md
@@ -31,10 +31,14 @@ dependencies {
 }
 ```
 
-Update Android SDK version if you did `react-native init`, we want to be on `26` or higher.
-* `compileSdkVersion 26`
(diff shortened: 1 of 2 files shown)


kristfal:
Looks good. Lgtm.

PR #1565, opened 2019-03-29
correct coordinate type on interface PointAnnotationProps to Array<number>

lauherk:
github issue: https://github.com/nitaliano/react-native-mapbox-gl/issues/1559

change coordinate type on interface PointAnnotationProps from Array string to Array number.


nitaliano:
Thanks for the PR!

PR #1593, opened 2019-04-25
Fixed mapping MGL enum values on RN 0.59

mattijsf:
Fixes:

https://github.com/nitaliano/react-native-mapbox-gl/issues/1498
https://github.com/nitaliano/react-native-mapbox-gl/issues/1589
https://github.com/nitaliano/react-native-mapbox-gl/issues/1110

I've noticed that since RN 0.59 react-native provides `NSNumber` values differently depending on the architecture of the platform, these are the container types `NSNumber` uses when providing `lineCap` (and other integer based values):

RN 0.57:
iPhone 5S: `kCFNumberSInt64Type`
iPhone SE: `kCFNumberSInt64Type`

RN 0.59:
iPhone 5S: `kCFNumberSInt64Type`
iPhone SE: **`kCFNumberFloat64Type`**

Normally this doesn't impose any problems but for some reason it produces the following error when mapping these numbers as a raw value into `MGLStyleValue` that is assigned to an `NS_ENUM` such as `lineCap` / `lineJoin` / `iconAnchor` etc:
> `Assertion failed: (it != std::end(MGLLineCap_names)), function toString, file /Users/distiller/project> /platform/darwin/src/MGLLineStyleLayer.mm`

This PR makes sure that all `enum` style properties values are treated as `NSUInteger` before wrapping them into `MGLStyleValue`.


kristfal:
Im good with merging this one.

PR #1631, opened 2019-05-28
Update type information

RossWilliams:
Modify hitbox property for ShapeSourceProps as optional to match react proptypes. Checked code to verify it does not assume hitbox to be non-null, and purpose of hitbox is to override the default of 44x44.

Modify CalloutProps to include title, used in Callout.js line 109.


kristfal:
This is ok. Also suggest you apply this fix to the new repository at https://github.com/react-native-mapbox-gl/maps

